PR #2, opened 2024-10-18
Add preprocess module (1st step)

roshankern:
This PR adds the first step in the preprocess module. We add the env, lib, rule, and script files necessary for this first step. We also add a `pyproject.toml` file to make the python file accessible.

@mat10d this is ready for review when you get a chance!

*Note*: I opted to make the first step output TSVs instead of pickles, but this is easy to revert if we would rather stick with pickles.


workflow/scripts/preprocess/extract_metadata_tile.py
mat10d, line 1: Super clean. Nice

workflow/rules/preprocess.smk
mat10d, line 1: Cool--so this is the main preprocess script

mat10d:
Ok, so just to summarize how this works--correct me if I'm wrong.

In example analysis, you run a bash script, which calls Snakefile, which calls the preprocess.smk, which calls the functions in the preprocess.py. Is that right?

mat10d (approved):

workflow/lib/preprocess/file_utils.py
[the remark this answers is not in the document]
roshankern: Yep! This is actually novel code for handling the new system with file paths stored in the config tsv files.

workflow/rules/preprocess.smk
roshankern: Yep!

roshankern:
> In example analysis, you run a bash script, which calls Snakefile, which calls the preprocess.smk, which calls the functions in the preprocess.py. Is that right?

I think you understand but I want to clarify the last part for this step in particular:
`example_analysis/run_analysis.sh` -> `workflow/Snakefile` -> `workflow/rules/preprocess.smk` -> `workflow/scripts/preprocess/extract_metadata_tile.py` -> helper code in `workflow/lib/preprocess/`

PR #3, opened 2024-10-24
Finish preprocessing migration

roshankern:
In this PR we:
- Finish the migration of the preprocessing module
- Lint/format code with Ruff
- Convert all docstrings to Google format

**Note**: I did not change the code to use the `nd2` package instead of `nd2reader` in this PR because it will require more testing.


README.md
```diff
@@ -25,3 +27,10 @@ sh run_analysis.sh
 ```
 
 *Note*: We do not include the data necessary for this example analysis in this repo as it is too large.
+
+
+## Contribution Notes
+
+- We use the [one sentence per line](https://nick.groenen.me/notes/one-sentence-per-line/) convention for markdown files.
```
mat10d, line 34: This is very good. Let's keep this up!

workflow/lib/preprocess/preprocess.py
```diff
@@ -1,61 +1,114 @@
...
         return combined_metadata
     else:
-        print(f"No valid ND2 files found in the provided list.")
+        print("No valid ND2 files found in the provided list.")
         return pd.DataFrame()  # Return an empty DataFrame if no files were processed
+
+
+def nd2_to_tif(file: str, channel_order_flip: bool = False) -> np.ndarray:
+    """
+    Converts a single ND2 file with one field of view and multiple channels to a multidimensional numpy array.
```
mat10d, line 82: So as we migrate, are we assuming this format? Are we ignoring the potential to use well based images?

workflow/lib/shared/illumination_correction.py
mat10d, line 1: Based on our discussion yesterday, I think you do have some logic to keep this file in shared. Might it make sense to have some sort of descriptor for each file? I think this was useful in the original codebase (main) that I had, but up to you. But this way, we could motivate the need to keep all the illumination correction functions in shared because it both calculates the correction and then applies it (you will add this later)

mat10d: I don't know if there is a format, (e.g. one sentence per line/ruff/google docstrings) for doing this, or if it isn't conventional to have that. But worth a thought.

workflow/rules/preprocess.smk
```diff
@@ -12,17 +12,104 @@ phenotype_samples_df = pd.read_csv(PHENOTYPE_SAMPLES_FP, sep="\t")
...
 print(f"Phenotype samples shape: {phenotype_samples_df.shape}")
 
+
 # Extract metadata for SBS images
 rule extract_metadata_sbs:
+    conda:
+        "../envs/preprocess.yml"
+    input:
+        lambda wildcards: get_sample_fps(
+            sbs_samples_df, well=wildcards.well, cycle=wildcards.cycle
```
mat10d, line 22: I haven't seen this before, but get_sample_fps is a function that returns wildcards?

workflow/scripts/preprocess/nd2_to_tif.py
```diff
@@ -0,0 +1,11 @@
...
+from lib.preprocess.preprocess import nd2_to_tif
+from skimage.io import imsave
+
+# convert the ND2 file to a TIF image array
+image_array = nd2_to_tif(
+    snakemake.input[0], channel_order_flip=snakemake.params.channel_order_flip
+)
+
+# save TIF image array to the output path
+imsave(snakemake.output[0], image_array)
```
mat10d, line 11: I think before you merge this, it would be good to test if the things that you are saving are compatible with Fiji. I would download Fiji and Napari

https://imagej.net/software/fiji/downloads
https://napari.org/dev/tutorials/fundamentals/installation.html

And see what this tiff format is compatible with

We will also need to check if things work well with downstream steps given that we are massively changing how we are saving images. I am assuming that you can compare the analysis on the few tiles that you did in dev with these. I am a bit worried about this, but also understand the importance of streamlining this

workflow/lib/preprocess/file_utils.py
```diff
@@ -1,28 +1,34 @@
-def get_sample_fps(samples_df, well=None, tile=None, cycle=None):
+import pandas as pd
+
+
+def get_sample_fps(
+    samples_df: pd.DataFrame, well: str = None, tile: int = None, cycle: int = None
+) -> list[str]:
     """
```
mat10d, line 7: I really like this approach. I have struggled heavily with wildcard usage downstream of preprocessing. Let's keep this up. Will be really useful for merge/hash (step 3)

mat10d (approved):
I would check Napari and Fiji compatibility, just to know if the output tiffs are compatible. But otherwise great start!

workflow/lib/preprocess/preprocess.py
roshankern: By format are you referring to the single ND2 file with one field of view and multiple channels? I'm not sure if we should assume this format and ignore the potential to use well-based images. Let's discuss over Zoom before we come to a conclusion on this.

workflow/lib/shared/illumination_correction.py
roshankern: > Might it make sense to have some sort of descriptor for each file? I think this was useful in the original codebase (main) that I had, but up to you.

Are you referring to the docstring at the top of each python file? I think you are right that this would be helpful (I think only really helpful in the lib files). I'll add this to the lib files.

workflow/scripts/preprocess/nd2_to_tif.py
roshankern: Makes sense. I'll check that these are compatible with Fiji/Napari. Definitely don't want to mess up the downstream steps.

workflow/lib/shared/illumination_correction.py
mat10d: Okay, I think that makes it easier for others to see what each file contains. Yes, referring to the docstring

workflow/scripts/preprocess/nd2_to_tif.py
mat10d: Sounds good

workflow/lib/shared/illumination_correction.py
roshankern: Done in https://github.com/cheeseman-lab/brieflow/pull/3/commits/654650b7f4e338192d1a6e5f9b3f5947f24c4e9a

workflow/scripts/preprocess/nd2_to_tif.py
roshankern: Confirmed I am able to view these converted tifs. Merging!

Issue #4, opened 2024-10-31
Preprocess getting tile metadata from file names


roshankern:
[This code](https://github.com/cheeseman-lab/brieflow/blob/0d52bd1b53957add81e025a4e52340b0f7041de5/workflow/lib/preprocess/preprocess.py#L12-L27) in preprocess derives tile numbers from file names. We want tile (and other image metadata) to be specified in config, so we should move away from this approach.

roshankern:
Fixed in #16 !

PR #5, opened 2024-10-31
Add file conventions

roshankern:
In this PR we:

- Add more conventions for the file types/names we will be using throughout the repo.
- Add a `get_filename` utility function that can be used with snakemake to generate file names from data location, info, type.
- Use `get_filename` with snakemake preprocessing calls.


README.md
```diff
@@ -31,6 +31,13 @@ sh run_analysis.sh
...
+We use [ruff](https://github.com/astral-sh/ruff) for linting and formatting code.
+
+We use the following conventions:
+- [One sentence per line](https://nick.groenen.me/notes/one-sentence-per-line/) convention for markdown files
+- [Google format](format) for function docstrings
+- [tsv](https://en.wikipedia.org/wiki/Tab-separated_values#:~:text=Tab%2Dseparated%20values%20(TSV),similar%20to%20comma%2Dseparated%20values.) file format for saving small dataframes that require easy readability
+- [hdf5](https://www.hdfgroup.org/solutions/hdf5/) file format for saving large dataframes
+- Data location information (well, tile, cycle, etc) + `__` + type of information (cell features, phenotype info, etc) + `.` + file type. 
+Data is stored in its respective analysis directories. 
+For example: `analysis_root/preprocess/ic_fields/sbs/A1_T1_C1__ic_field.tiff`
```
mat10d, line 43: The file naming convention is great. Is the analysis directory fixed in this way?

workflow/Snakefile
mat10d, line 1: This is definitely cleaner

workflow/lib/preprocess/preprocess.py
```diff
@@ -6,7 +6,7 @@
 import numpy as np
 from nd2reader import ND2Reader
 
-from preprocess.file_utils import extract_tile_from_filename
+from shared.file_utils import extract_tile_from_filename
```
mat10d, line 9: Makes sense to move to shared as you're scaling this across all aspects of the pipeline, right?

mat10d (commented):
Looks good to me to merge. The path to where the images reside may require more thinking, but I think the example you are putting in the readme is just an example?

README.md
roshankern: That is an actual path yes. I'm open to more discussion about the organization of processing directories. In this case it made sense to me to have module (preprocess) -> subprocess (ic_fields) -> data type (sbs or ph; this is only a relevant distinction for preprocessing module). 

Basically with this file convention I was saying that the file names consistently contain:
1) Data location (well, tile, cycle)
2) Information type (ic_field, aligned_image, etc)
3) File type (tiff, tsv, etc)
And all other information (module, subprocess, sbs vs ph) gets stored as directories. Let me know if my logic is off here.

workflow/lib/preprocess/preprocess.py
roshankern: I actually didn't mean to move this to `shared` (we only use it in preprocessing) and am moving it back in the next PR. But, I will keep a `shared/file_utils.py` file because we definitely want `get_filename` in the shared utils. 

I don't think the location of this `extract_tile_from_filename` function is important though because I actually want to remove it entirely soon. See #4

PR #9, opened 2024-11-07
change to nd2 package for read in

mat10d:
Not much of a change, but eliminated nd2reader and added nd2 as the default file read in. Some changes to the preprocessing scripts. Also added functionality to specify the number of z dimensions to flatten on, this will make metadata generation easier.

### DIFF
--- a/workflow/envs/preprocess.yml
+++ b/workflow/envs/preprocess.yml
@@ -3,6 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python=3.11
-  - nd2reader
+  - nd2
+  - xarray
   - scikit-image
   - joblib
--- a/workflow/lib/preprocess/preprocess.py
+++ b/workflow/lib/preprocess/preprocess.py
@@ -78,8 +78,7 @@ def extract_metadata_tile(files: list[str], z_interval: int = 4, verbose: bool =
 
 
 def nd2_to_tiff(file: str, channel_order_flip: bool = False, verbose: bool = False) -> np.ndarray:
-    """Converts a single ND2 file with one field of view and multiple channels to a multidimensional numpy array, 
-    ensuring the structure is CYX.
+    """Converts a single ND2 file with one field of view and multiple channels to a multidimensional numpy array, ensuring the structure is CYX.
 
     Args:
         file (str): Path to the ND2 file.
--- a/workflow/lib/preprocess/preprocess.py
+++ b/workflow/lib/preprocess/preprocess.py
@@ -2,94 +2,111 @@
 
 import pandas as pd
 import numpy as np
-from nd2reader import ND2Reader
+import nd2
 
 from lib.preprocess.file_utils import extract_tile_from_filename
 
 
-def extract_metadata_tile(files: list[str]) -> pd.DataFrame:
+def extract_metadata_tile(files: list[str], z_interval: int = 4, verbose: bool = False) -> pd.DataFrame:
     """Extracts metadata from a list of ND2 files.
 
     Args:
         files (list[str]): List of file paths pointing to ND2 files.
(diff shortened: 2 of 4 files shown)


workflow/envs/preprocess.yml
```diff
@@ -3,6 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python=3.11
-  - nd2reader
+  - nd2
+  - xarray
```
roshankern, line 7: Do we need to specify xarray as a dependency here? It seems like nd2 uses this (at least I see reference to it later) but if we don't need to import it anywhere I would think we don't need to include it as a dependency ourselves.

workflow/lib/preprocess/preprocess.py
```diff
@@ -2,94 +2,111 @@
 
 import pandas as pd
 import numpy as np
-from nd2reader import ND2Reader
+import nd2
```
roshankern, line 5: Supported package!

workflow/lib/preprocess/preprocess.py
```diff
@@ -2,94 +2,111 @@
...
+    
+    print("No valid ND2 files found in the provided list.")
+    return pd.DataFrame()
 
 
-def nd2_to_tiff(file: str, channel_order_flip: bool = False) -> np.ndarray:
-    """Converts a single ND2 file with one field of view and multiple channels to a multidimensional numpy array.
+def nd2_to_tiff(file: str, channel_order_flip: bool = False, verbose: bool = False) -> np.ndarray:
+    """Converts a single ND2 file with one field of view and multiple channels to a multidimensional numpy array, 
+    ensuring the structure is CYX.
```
roshankern, line 82:
```suggestion
    """Converts a single ND2 file with one field of view and multiple channels to a multidimensional numpy array, ensuring the structure is CYX.
```

I am following this convention for one line at the top of the docstring because it is enforced by ruff. I would recommend running ruff check in Brieflow/ before filing PR to make sure ruff doesn't complain about anything.

workflow/lib/preprocess/preprocess.py
```diff
@@ -2,94 +2,111 @@
...
+    image_array = image.transpose('C', 'Y', 'X').values
+
+    # Flip channel order on C axis if requested
+    if channel_order_flip:
+        image_array = np.flip(image_array, axis=0)  # axis 0 is the C axis
+
+    if verbose:
+        print(f"Final dimensions: {image_array.shape}")
+    
+    return image_array.astype(np.uint16)
```
roshankern, line 112: Blank line at the end of the file here. Using ruff format in Brieflow/ will help avoid these little things as well. You can also make python files [autoformat](https://stackoverflow.com/questions/50606758/vscode-how-do-you-autoformat-on-save) on save in vscode.

roshankern (approved):
LGTM!

Just some small comments.

roshankern:
One more note @mat10d 

Try to make branch names clear as we will likely have lots of them in the future!
The name for this PR was great, maybe something like `implement_nd2_loading` for the branch name.

workflow/envs/preprocess.yml
mat10d: nd2 will import it as default, but to read the metadata, we need xarray to be installed in this conda env. So yes, we need it here

workflow/lib/preprocess/preprocess.py
mat10d: Fixed, thanks for the heads up.

workflow/lib/preprocess/preprocess.py
mat10d: Also changed. Running ruff is easy. Should we include it as a requirement to brieflow or is that not standard? I had to install in another conda env.

mat10d:
My bad, I thought I was going to make more changes but smaller branches is defo the move. Will make better names going forward.

PR #10, opened 2024-11-07
Migrate sbs eval + Create rulegraph

roshankern:
In this PR we:
- migrate the sbs eval code from OPS into the snakemake pipeline
- create a rulegraph for the current components of the snakefile
Note: We cannot use `print()` from snakefiles if we want to create rulegraphs because this causes errors.


example_analysis/config/config.yml
```diff
@@ -12,19 +12,17 @@
...
-#   'output_fp' where the results of that section are put.
 
 
 # General options
 all:
-  root_fp: "/Users/roshankern/Desktop/Github/BrieFlow/example_analysis/analysis_root"
-  output_fp: "brieflow_output/"
+  root_fp: "analysis_root/"
 
 # Preprocessing
```
mat10d, line 21: These are huge. Allow us to both work on it. Thanks!

example_analysis/run_analysis.sh
```diff
@@ -3,6 +3,10 @@
 # Set the path to the main Snakefile and the config file
 SNAKEFILE="../workflow/Snakefile"
 CONFIG_FILE="config/config.yml"
+RULEGRAPH_FILE="../images/brieflow_rulegraph.png"
+
+# Generate a rulegraph of the Snakefile
+# snakemake --snakefile "$SNAKEFILE" --configfile "$CONFIG_FILE" --rulegraph | dot -Gdpi=100 -Tpng -o "$RULEGRAPH_FILE"
```
mat10d, line 9: Not working? Why commented out?

workflow/scripts/sbs_process/eval.py
```diff
@@ -0,0 +1,91 @@
...
+    plot_gene_symbol_histogram,
+)
+
+# Read barcodes
+df_design = pd.read_csv(snakemake.params.df_design_path, sep="\t")
+df_pool = df_design.query("dialout==[0,1]").drop_duplicates("sgRNA")
+df_pool["prefix"] = df_pool.apply(lambda x: x.sgRNA[: x.prefix_length], axis=1)
+barcodes = df_pool["prefix"]
+
+# Concatenate files
```
mat10d, line 20: Does it make sense to leave this in eval? Should this just be part of our combine_* rules? 

The well based combine_* aren't really used for anything...

mat10d: This would also make eval truly an eval script...

workflow/scripts/sbs_process/eval.py
```diff
@@ -0,0 +1,91 @@
...
+df_summary_any.to_csv(snakemake.output[5], index=False, sep="\t")
+plt.gcf().savefig(snakemake.output[6])
+plt.close()
+
+outliers = plot_reads_per_cell_histogram(cells, x_cutoff=20)
+plt.savefig(snakemake.output[7])
+plt.close()
+
+outliers = plot_gene_symbol_histogram(cells, x_cutoff=30)
+plt.gcf().savefig(snakemake.output[8])
```
mat10d, line 70: So it seems for snakemake compatibility, actually having outputted figs (by functions like plot_gene_symbol_histogram) makes a lot of sense, that way it can be saved to the path specified in the snakemake file? Is this accurate? We should keep this standardized throughout.

workflow/lib/sbs_process/eval.py
```diff
@@ -0,0 +1,584 @@
...
+import numpy as np
+import seaborn as sns
+import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
+
+
+# helper function to load and concatenate hdfs
+# helper function to load and concatenate HDFs
+def load_and_concatenate_hdfs(hdf_files):
+    """Load and concatenate HDF files from a provided list into a single DataFrame.
```
mat10d, line 15: Maybe this is moved. See below comments.

workflow/scripts/sbs_process/eval.py
```diff
@@ -0,0 +1,91 @@
...
+
+plot_mapping_vs_threshold(reads, barcodes, "peak")
+plt.gcf().savefig(snakemake.output[0])
+plt.close()
+
+plot_mapping_vs_threshold(reads, barcodes, "Q_min")
+plt.gcf().savefig(snakemake.output[1])
+plt.close()
+
+plot_read_mapping_heatmap(reads, barcodes, shape="6W_sbs")
```
mat10d, line 35: Should this look like all the other heatmap functions in the code?

workflow/scripts/sbs_process/eval.py
```diff
@@ -0,0 +1,91 @@
...
+plt.gcf().savefig(snakemake.output[2])
+plt.close()
+
+df_summary_one, _ = plot_cell_mapping_heatmap(
+    cells,
+    minimal_phenotype_info,
+    barcodes,
+    mapping_to="one",
+    mapping_strategy="gene_symbols",
+    shape="6W_sbs",
```
mat10d, line 45: Please note that in the phenotyping step, the shape may incorrectly be 6W_sbs. I will remember this and bring it up later as well. Good here.

workflow/scripts/sbs_process/eval.py
```diff
@@ -0,0 +1,91 @@
...
+cells["mapped_single_gene"] = cells.apply(
+    lambda x: (
+        True
+        if (pd.notnull(x.gene_symbol_0) & pd.isnull(x.gene_symbol_1))
+        | (x.gene_symbol_0 == x.gene_symbol_1)
+        else False
+    ),
+    axis=1,
+)
+print(cells.mapped_single_gene.value_counts())
```
mat10d, line 84: This section needs to be more informative. I may suggest that we have a function that does this:
0. How many cells are segmented?
1. How many cells have a single barcode?
2. How many cells have > 1 barcode?
 (for 1 and 2, what are these percentages?) This txt file needs to be a super clear output. Maybe ask in the ops channel what outputs others would expect.

mat10d: In this, the outputs that are generated here are useful as a first step--can you check that they would output what is coming out of the plots? We need to make sure these align

mat10d (changes requested):
Overall, looks very good. The one thing that I would change now, looking at things, is that the well based combines that are happening in the first part of sbs_process are a bit irrelevant, and actually what we want is the final fully merged hdf. Furthermore, the fully merged hdf is not really part of any evaluation step--so maybe move that out of eval and condense that into one rule?

workflow/lib/sbs_process/eval.py
```diff
@@ -0,0 +1,584 @@
...
+        # Plot heatmap
+        axes = plot_plate_heatmap(df_summary, shape=shape, plate=plate, **kwargs)
+        return axes
+    elif return_summary:
+        return df_summary
+    else:
+        return None
+
+
+def plot_plate_heatmap(
```
mat10d, line 187: This is a shared function. It is used in other steps as well!

example_analysis/run_analysis.sh
roshankern: It works but takes a little bit of extra computation so I want to keep it commented out while we are developing Brieflow and are constantly running this script.

I added a note about this in https://github.com/cheeseman-lab/brieflow/pull/10/commits/958abc37b07cdcdc85d60e282f0ca6106201f43f

workflow/lib/sbs_process/eval.py
roshankern: Thanks for pointing this out! Done in https://github.com/cheeseman-lab/brieflow/pull/10/commits/f4d9c741ca2cdb6a4951b6a1c3e4a24e6035e89e

workflow/scripts/sbs_process/eval.py
roshankern: Yep, outputted figs is the move. Good for consistency with other scripts because we always load and save files outside of lib functions.

Made all of these helper functions return figures in https://github.com/cheeseman-lab/brieflow/pull/10/commits/f460dd8242afd36458b3c8843873826652a760b2

workflow/scripts/sbs_process/eval.py
roshankern: Is it different? I updated all of these heatmap function to be used like 

```
fig = plot_read_mapping_heatmap(reads, barcodes, shape="6W_sbs")
fig.savefig(snakemake.output[2])
```

in https://github.com/cheeseman-lab/brieflow/pull/10/commits/f460dd8242afd36458b3c8843873826652a760b2

workflow/scripts/sbs_process/eval.py
roshankern: Yep this makes sense. Done in https://github.com/cheeseman-lab/brieflow/pull/10/commits/c1df54de3a10686c63613494f2bef3c8d1802b5e

workflow/lib/sbs_process/eval.py
roshankern: Done in https://github.com/cheeseman-lab/brieflow/pull/10/commits/4052c72e7e10a02317e6ad41f35d23ba0a13a276

workflow/scripts/sbs_process/eval.py
roshankern: Done in the last couple of commits! Hopefully I covered everything necessary to merge! See my comment below for more information.

roshankern:
@mat10d This is ready for re-review. I have broken the eval step into:

`eval_segmentation`:
- cell density stats: tsv and plate heatmap showing density of segmented cells
- segmentation overview stats: all info from segment_cells return_counts including initial_nuclei, initial_cells, after_edge_removal_nuclei, after_edge_removal_cells, final_cells, final_nuclei. Info averaged across wells

`eval_mapping`:
- everything we had before except clarified that plate heatmaps were showing gene symbol mapping statistics, not barcode mapping statistics
- mapping overview stats: df with info for total_cells__count, 1_barcode_cells__count, 1_barcode_cells__percent, 1_or_more_barcodes__count, 1_or_more_barcodes__percent, 1_gene_cells__count, 1_gene_cells__percent, 1_or_more_genes__count, 1_or_more_genes__percent. Info averaged across wells

I tried to include everything we discussed but let me know if I missed anything!

workflow/scripts/sbs_process/eval.py
mat10d: Looks good to merge

mat10d:
Go ahead and merge. Let's discuss outputs at our catch-up

PR #12, opened 2024-11-14
Add Configuration + Workflow Setup

roshankern:
In this PR we restructure the example analysis of Brieflow so that users can see how to configure parameters in between workflow steps.

The general idea is that users configure parameters with notebooks then run the Snakemake-based workflows with an sh file. Please see new Brieflow Readme for more details!


mat10d (approved):
Overall, I think this is good to go. I reviewed much of this with the subsequent PR. I think it's really smart that you're making things abundantly clear in the notebooks about what parameters are fixed, and which can be changed.

PR #13, opened 2024-11-19
Separate Target Definitions

roshankern:
In this PR we move all output/target definitions to separate files. The new process for defining and using outputs is as follows:

1) For a process, create a snakemake file in targets like `targets/preprocess.smk`
2) Create an outputs dictionary that maps a rule name to a list of file path outputs. Ex:
```
PREPROCESS_OUTPUTS = {
    "extract_metadata_sbs": [
        PREPROCESS_FP
        / "metadata"
        / "sbs"
        / get_filename({"well": "{well}", "cycle": "{cycle}"}, "metadata", "tsv"),
    ],
}
```
3) Define output types as `None`, `temp`, `protected`, etc
4) Map outputs to their types with `lib.shared.target_utils.map_outputs`
5) Use these outputs in a Snakemake rule. Ex:
```
# Apply Laplacian-of-Gaussian filter to all channels
rule log_filter:
    conda:
        "../envs/sbs_process.yml"
    input:
        SBS_PROCESS_OUTPUTS["align"],
    output:
        SBS_PROCESS_OUTPUTS_MAPPED["log_filter"],
    params:
        skip_index=0,
    script:
        "../scripts/sbs_process/log_filter.py"
```

@mat10d this is ready for review! Let me know if you have any questions!


README.md
```diff
@@ -86,8 +114,14 @@ sh run_analysis.sh
 
 ## Contribution Notes
 
+
+### Dev Tools
```
mat10d, line 118: Good to separate, it was getting long

workflow/Snakefile
```diff
@@ -20,198 +20,30 @@ SBS_WELLS = list(sbs_samples_df["well"].unique())
...
 
+include: "targets/preprocess.smk"
 include: "rules/preprocess.smk"
 
 
+ALL_TARGETS = PREPROCESS_TARGETS_ALL
+
 if "sbs_process" in config:
 
+    include: "targets/sbs_process.smk"
```
mat10d, line 40: It makes sense overall. Targets and rules are separated out. I think this looks cleaner.

workflow/Snakefile
```diff
@@ -20,198 +20,30 @@ SBS_WELLS = list(sbs_samples_df["well"].unique())
...
     include: "rules/sbs_process.smk"
 
-
-PREPROCESS_FP = ROOT_FP / "preprocess"
-SBS_PROCESS_FP = ROOT_FP / "sbs_process"
+    ALL_TARGETS += SBS_PROCESS_TARGETS_ALL
```
mat10d, line 43: This way we can also add them sequentially to the overall snakefile. Good.

workflow/lib/preprocess/preprocess.py
```diff
@@ -90,22 +96,22 @@ def nd2_to_tiff(file: str, channel_order_flip: bool = False, verbose: bool = Fal
...
-    image_array = image.transpose('C', 'Y', 'X').values
+    image_array = image.transpose("C", "Y", "X").values
 
     # Flip channel order on C axis if requested
     if channel_order_flip:
         image_array = np.flip(image_array, axis=0)  # axis 0 is the C axis
 
     if verbose:
         print(f"Final dimensions: {image_array.shape}")
-
```
mat10d, line 110: There aren't any noticeable changes here. I think I just need to retest it. I don't want to revert back to nd2Reader. you don't need this data anyway for the time being...

One thing to think about--the preprocess takes each nd2 file and creates a tiff. However, it takes a group of nd2 files to create the metadata. This is a bit annoying, and also requires us to have different functions for processing metadata in a tile or well based format. It may make more sense to have one metadata file per each nd2 file, and then have a combine_metadata step. This may also assist with debugging, given your error message. Let's discuss this later.

workflow/rules/sbs_process.smk
```diff
@@ -1,30 +1,19 @@
...
-            / "images"
-            / "sbs"
-            / get_filename(
-                {"well": wildcards.well, "tile": wildcards.tile, "cycle": "{cycle}"},
-                "image",
-                "tiff",
-            ),
-            cycle=SBS_CYCLES,
+        lambda wildcards: output_to_input(
+            PREPROCESS_OUTPUTS["convert_sbs"],
```
mat10d, line 11: Ok I see. So both the target and actual rules reference these set of outputs?

This eliminates a ton of redundancy.

workflow/scripts/preprocess/extract_metadata_tile.py
```diff
@@ -1,5 +1,7 @@
 from lib.preprocess.preprocess import extract_metadata_tile
 
+print(snakemake.input)
```
mat10d, line 3: Is this here because you are trying to debug?

mat10d (approved):
This looks awesome. All for merging this.

workflow/lib/preprocess/preprocess.py
roshankern: > you don't need this data anyway for the time being...

Yep! 

> It may make more sense to have one metadata file per each nd2 file, and then have a combine_metadata step. This may also assist with debugging, given your error message. Let's discuss this later.

This definitely sounds like a good idea.

workflow/rules/sbs_process.smk
roshankern: > So both the target and actual rules reference these set of outputs?

Exactly. We only need to specify the outputs and their mapping types then can base all inputs/outputs/targets on this.

workflow/scripts/preprocess/extract_metadata_tile.py
roshankern: Yep, good catch! Removed unnecessary print statements in https://github.com/cheeseman-lab/brieflow/pull/13/commits/583d15c08147212b3b1ddf40c9acbf1a76484f46

PR #15, opened 2024-11-22
Fix ND2 metadata extraction

mat10d:


roshankern:
Looks good, thanks for fixing!

PR #16, opened 2024-11-23
Refactor extract metadata

roshankern:
In this PR we factor the ND2 metadata extraction to extract individual metadata statistics for each ND2 files then combining these metadata stats across tiles into one TSV file. We also resolve the issue mentioned in #4.


workflow/lib/preprocess/file_utils.py
```diff
@@ -34,18 +32,3 @@ def get_sample_fps(
 
     # Return the list of file paths as Path objects
     return filtered_df["sample_fp"].tolist()
```
mat10d, line 34: I got a comment from Sarah, who I've connected you to, that the discrete file naming approach is a pain for instances where the file structure changes. So this approach will be much better.

workflow/rules/preprocess.smk
```diff
@@ -11,14 +11,34 @@ rule extract_metadata_sbs:
...
         PREPROCESS_OUTPUTS_MAPPED["extract_metadata_sbs"],
     params:
         z_interval=None,
+        tile=lambda wildcards: wildcards.tile,
+    script:
+        "../scripts/preprocess/extract_tile_metadata.py"
+
+
+# Extract metadata for SBS images
+rule combine_metadata_sbs:
```
mat10d, line 27: Great!

mat10d (commented):
Simple and clean. Looks good to merge

PR #18, opened 2024-11-26
Adding HPC capabilites to brieflow

mat10d:
This workflow develops HPC capabilites for brieflow. This is optimized for use on the wibr cluster--through a bespoke cluster.yaml file in the slurm subfolder. I would imagine others would redevelop some aspects of this, but the general outline would remain the same.

Given the information we saw for sunbeam, I sought to make this fully slurm compatible. There are two general approaches to running on slurm:

1. A slurm job that calls multiple slurm jobs
2. A bash file that calls multiple slurm jobs

For a variety of reasons, in this PR we are doing option 1. This allows for the separate development of slurm files outside of the current workflow.

There isn't much to review here, but it works, and any renaming/organization conventions that you think would be useful would be much appreciated, as well as changes in the README if you'd like.


brieflow_workflows_env.yml
```diff
@@ -13,3 +13,4 @@ dependencies:
   - numpy
   - pip:
       - -e .
+      - snakemake-executor-plugin-slurm
```
roshankern, line 16: It looks like there is an actively maintained [conda package](https://anaconda.org/bioconda/snakemake-executor-plugin-slurm) for this. Are we able to move to that package for consistent package management? I've also been reading in the snakemake docs that conda is the preferable package manager where possible.

Definitely worth making sure that the environment works if we do end up going with the conda verison though.

example_analysis/1.run_preprocessing_slurm.sh
```diff
@@ -0,0 +1,27 @@
...
+# Activate conda environment (adjust path as needed)
+source ~/.bashrc
+conda activate brieflow_workflows
+
+# Run Snakemake
+snakemake --use-conda --executor slurm \
+    --workflow-profile "$PROFILE_FILE_PATH" \
+    --snakefile "${BRIEFLOW_PATH}workflow/Snakefile" \
+    --configfile "$CONFIG_FILE_PATH" \
+    --until all_preprocess
```
roshankern, line 27:
```suggestion
    --until all_preprocess
    
```

Following convention to [end files with new line](https://dev.to/documendous/why-you-should-end-your-source-files-with-a-new-line-156g#:~:text=One%20common%20rule%20is%20to,codebase%20remains%20clean%20and%20maintainable.)

example_analysis/3.run_sbs_process_slurm.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Activate conda environment (adjust path as needed)
+source ~/.bashrc
+conda activate brieflow_workflows
+
+# Run only the preprocess rules
+snakemake --use-conda --executor slurm \
+    --workflow-profile "$PROFILE_FILE_PATH" \
+    --snakefile "${BRIEFLOW_PATH}workflow/Snakefile" \
+    --configfile "$CONFIG_FILE_PATH"
```
roshankern, line 26:
```suggestion
    --configfile "$CONFIG_FILE_PATH"
    
```

example_analysis/run_entire_analysis_slurm.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Generate a rulegraph of the Snakefile
+# NOTE: Uncomment when needed, takes extra computation
+# snakemake --snakefile "$SNAKEFILE" --configfile "$CONFIG_FILE" --rulegraph | dot -Gdpi=100 -Tpng -o "$RULEGRAPH_FILE"
+
+# Run Snakemake with the specified Snakefile and config file
+snakemake --use-conda --executor slurm \
+    --workflow-profile "$PROFILE_FILE_PATH" \
+    --snakefile "${BRIEFLOW_PATH}workflow/Snakefile" \
+    --configfile "$CONFIG_FILE_PATH"
```
roshankern, line 31:
```suggestion
    --configfile "$CONFIG_FILE_PATH"
    
```

example_analysis/slurm/config.yaml
```diff
@@ -0,0 +1,60 @@
...
+        mem_mb: 200
+    extract_sbs_info:
+        mem_mb: 200
+    combine_reads:
+        mem_mb: 20000
+    combine_cells:
+        mem_mb: 20000
+    combine_sbs_info:
+        mem_mb: 20000
+    # phenotype_process
```
roshankern, line 60:
```suggestion
    # phenotype_process
    
```

roshankern (changes requested):
Requesting changes because of my hope that we can stick with Conda for package management. Let me know if you find the slurm plugin is only available for our use case through pip.

roshankern:
One more thought I want to add while this PR is going through.

The example analysis on this repo will probably start to get cluttered as we add more modules and functionality (tile vs well based, local vs on HPC, etc). Given that this repo is supposed to be a software repo for the Brieflow tool, we should move away from having analysis in this repo long term if possible.

To align with the [Way Lab's GitHub strategy](https://wayscience.github.io/playbooks/github_strategy.html) that we are following, we should have Brieflow as a software repo only and another repo that shows how to use Brieflow for useful OPS insights.

Of course, this is a long term change but I want to get it on the radar as the example analysis starts to get more cluttered.

mat10d:
I don't think that the Way lab strategy applies to what we are doing here, but it will apply long term to what we are doing. Once we have developed this approach, we should release it with the example_analysis directory as is. When we actually analyze a screen, I agree, we should have an analysis repo. If we are so confident that brieflow is working effectively, we could make it pip installable and have the example_analysis in another repo that would be cloned and be the setup of another analysis, but I think we are a ways from this.

example_analysis/1.run_preprocessing_slurm.sh
mat10d: Changed in latest [commit](https://github.com/cheeseman-lab/brieflow/pull/18/commits/485b5e89ea72815b1fdb0a3c596c547425a9c232)

example_analysis/3.run_sbs_process_slurm.sh
mat10d: Changed in latest [commit](https://github.com/cheeseman-lab/brieflow/pull/18/commits/485b5e89ea72815b1fdb0a3c596c547425a9c232)

example_analysis/run_entire_analysis_slurm.sh
mat10d: Changed in latest [commit](https://github.com/cheeseman-lab/brieflow/pull/18/commits/485b5e89ea72815b1fdb0a3c596c547425a9c232)

example_analysis/slurm/config.yaml
mat10d: Changed in latest [commit](https://github.com/cheeseman-lab/brieflow/pull/18/commits/485b5e89ea72815b1fdb0a3c596c547425a9c232)

PR #19, opened 2024-12-02
Add local phenotype process

roshankern:
In this PR we finish migration of the phenotype process code. This includes

- Remaining library code necessary for phenotype process
- Env, snakemake rule, and targets for phenotype process
- Evaluation code for segmentation and features (integrated into snakemake rule for this module)

**Note:** Because Brieflow can operate with many modules as part of the same snakemake workflow (ie sbs and phenotype processes as part of same snakemake workflow), it is necessary to have unique rule names across these modules. Thus, `_sbs` and `_phenotype` suffixes were added to relevant shared rules. Ex: `segment_cellpose` -> `segment_cellpose_sbs` and `segment_cellpose_phenotype`. @mat10d We need to change references to these rules with changed names (ie in Slurm resources config). I left this refactor for another PR as I cannot test these changes.


example_analysis/3.configure_phenotype_process_params.ipynb
```diff
@@ -188,29 +188,29 @@
     "### Segmentation\n",
     "\n",
     "- `DAPI_INDEX`: The index of the DAPI channel in your image stack (0-indexed). Typically 0 for the DAPI channel from sequencing.\n",
-    "- `CYTO_CHANNEL`: The index of the cytoplasmic channel in your image stack (0-indexed). Typically the channel that best outlines cell boundaries, usually base C.\n",
```
mat10d, line 191: Good to rename this because it is an index. But would it make sense to have it as a channel, and then call that from the list of provided channels in the notebook/snakemake? That may make it easier for users to understand. Up to you though.

example_analysis/3.configure_phenotype_process_params.ipynb
```diff
@@ -188,29 +188,29 @@
...
+    "- `CYTO_INDEX`: The index of the cytoplasmic channel in your image stack (0-indexed). Typically the channel that best outlines cell boundaries, usually base C.\n",
```
mat10d, line 191: In this case, it is not a base, so the description here is inaccurate. However, my previous comment stands also for the SBS segmentation.

example_analysis/4.run_sbs_phenotype_processes.sh
```diff
@@ -4,7 +4,16 @@
...
+# Run the SBS process rules
+snakemake --use-conda --cores all \
+    --snakefile "${BRIEFLOW_PATH}workflow/Snakefile" \
+    --configfile "$CONFIG_FILE_PATH" \
+    --until all_sbs_process
+
+# Run the phenotype process rules
 snakemake --use-conda --cores all \
     --snakefile "${BRIEFLOW_PATH}workflow/Snakefile" \
-    --configfile "$CONFIG_FILE_PATH" \
```
mat10d, line 10: Please, can you ensure that all the processes.sh files always have an until flag? Even if it is the last module you have worked on? For the future.

workflow/Snakefile
```diff
@@ -43,6 +46,14 @@ if "sbs_process" in config:
...
 
+if "phenotype_process" in config:
+
+    include: "targets/phenotype_process.smk"
+    include: "rules/phenotype_process.smk"
+
+    ALL_TARGETS += PHENOTYPE_PROCESS_TARGETS_ALL
+
+
 # Define the target files for the workflow
```
mat10d, line 57: Does this also show up in RED (as if it is a syntax error) in your vscode when you have this file open? It still works, but this is a bit frustrating.

workflow/lib/external/cp_emulator.py
```diff
@@ -40,151 +40,6 @@
 )
 # from scipy.ndimage import map_coordinates # only required for granularity spectrum, which is currently unused
 
-######################################################################################################################################
-
-## CELLPROFILER SEGMENTATION FUNCTIONS (UNUSED)
```
mat10d, line 45: Good to remove.

workflow/scripts/phenotype_process/eval_features.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Load SBS processing files
+phenotype_cp_min = pd.read_hdf(snakemake.input[0])
+
+# Generate and save feature heatmaps
+min_feature_names = [col for col in phenotype_cp_min.columns if col.endswith("_min")]
+for feature_name in min_feature_names:
+    # Generate feature heatmap evaluation
+    df_summary_one, fig = plot_feature_heatmap(
+        phenotype_cp_min, feature=feature_name, shape="6W_sbs", return_summary=True
```
mat10d, line 14: Should be ph layout

workflow/scripts/phenotype_process/apply_ic_field_phenotype.py
```diff
@@ -0,0 +1,16 @@
+from skimage.io import imsave
+
+from lib.shared.file_utils import read_stack
+from lib.shared.illumination_correction import apply_ic_field
+
+# load raw image data
+raw_image_data = read_stack(snakemake.input[0])
```
mat10d, line 7: Same goes for my comments in the other PR. Seems like read_stack is only used in these two instances.

workflow/lib/shared/features.py
mat10d, line 1: Here, also, it seems like there are a lot of features that aren't showing up in the final set of feature tables. So maybe also worth a cleanup.

mat10d (changes requested):
Lots of comments, the main one that I think we will need to do at some point, and now is a good point is: for all the feature extraction functions, can we get a clear idea of how things are working and remove redundant code that is being used? This may take you 1/2 days of a deep dive and some minor cleanup, but also, a PR isn't the best way for us to interact on this. I think I will personally have to dig around and try to clean things up, but having a roadmap on how to do this would be really beneficial. Let's discuss at the call.

example_analysis/4.run_sbs_phenotype_processes.sh
roshankern: Yep! Sorry I missed this

workflow/Snakefile
roshankern: No it does not. I think this must be a VSCode configuration thing :(

workflow/scripts/phenotype_process/apply_ic_field_phenotype.py
roshankern: Resolved in #20

example_analysis/3.configure_phenotype_process_params.ipynb
roshankern: Good point! Done in https://github.com/cheeseman-lab/brieflow/pull/19/commits/252b80b92623b56e5f1c244306d956e31b5dd2e3

workflow/scripts/phenotype_process/eval_features.py
roshankern: Done in https://github.com/cheeseman-lab/brieflow/pull/19/commits/4f9b5573489c309e6959f983cc37532b05cba341

workflow/lib/phenotype_process/extract_phenotype_cp_multichannel.py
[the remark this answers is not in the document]
roshankern: > make sure this works on the blank images

Done in #20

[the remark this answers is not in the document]
roshankern: Many of these changes will be significantly easier once we transition to measure_cp, so for now I'm opting to make the easier changes below. This way we have short term functionality and can leave the heavier lifts for when we transition to the new package. See below!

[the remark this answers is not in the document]
roshankern: > There should be a way of printing the feature dictionary before it is applied. Can we do this?

For now, I am printing the features that are being extracted in the configure phenotype process notebook. I figure this is a good way for users to see what features are being extracted during testing.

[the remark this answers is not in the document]
roshankern: > regroup and rename the emulator functions
We reviewed this during our sync. For our short-term fix I have created a `feature_extraction.py` utility file with `extract_features` and `extract_features_bare`. Done in https://github.com/cheeseman-lab/brieflow/pull/19/commits/7525791f4148e8390b83746b1507eebe019135ec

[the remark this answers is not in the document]
roshankern: > see which sets of functions generally fail on which mask types
> make a quick summary of this in a word document

I will do this but want to merge this PR to unblock your changes!

workflow/lib/shared/features.py
roshankern: Okay definitely need to check this out but prioritizing merging this to unblock your changes. Thank you for highlighting!

mat10d (approved):
Looks good to merge!

Issue #11, opened 2024-11-13
Using external tifffile code


roshankern:
We use `Tifffile` for `read_stack` with the code for this class in [workflow/lib/shared/external/tifffile.py](https://github.com/cheeseman-lab/brieflow/blob/main/workflow/lib/shared/external/tifffile.py).

This code from Christoph Gohlke does exist as an actively maintained [python package](https://github.com/cgohlke/tifffile/tree/master) so we should probably use this package instead.

@mat10d has informed me that we don't know exactly why a separate file was implemented, so we should make sure that using this package has no unintended consequences.

roshankern:
Fixed in #20!

PR #20, opened 2024-12-03
Use tifffile package

roshankern:
In this PR, we move to the package version of [tifffile](https://github.com/cgohlke/tifffile/tree/master) (instead of copy-pasting code). The imread function of this package is used to load images throughout Brieflow.

Addresses #11 


mat10d (changes requested):
This is great, that you've made these changes. However, this suggests that in extract bases, we just need to squeeze the data and remove the 0th dimension for it ([here](https://github.com/cheeseman-lab/brieflow/blob/958ef16d4ff2b2b16183b6190cacdc8ea7acce1f/workflow/scripts/sbs_process/extract_bases.py#L12C1-L13C1)) So could we use imread, and then squeeze out the 0th dimension (if of size 1) in extract bases directly?

roshankern:
Made some changes to this PR to do the following:

- Use tifffile `imread`, `imwrite` instead of skimage functions. It turns out these work perfectly for all of our functions and don't need the `read_stack` wrapper.
- Make the entire pipeline empty-image compatible. There are various small changes throughout the PR that make the tests for this work.

mat10d (approved):
Ok this looks great--it worked including the merges of all min_cp data (the combine rules)?

If so, please merge. Thanks for doing this, I know it was a headache.

PR #24, opened 2024-12-10
Add sample pattern LLM prompt

roshankern:
This short PR adds an LLM prompt to the `0.configure_preprocess_params` notebook to derive the sample regex patterns from file names and descriptions.


example_analysis/0.configure_preprocess_params.ipynb
```diff
@@ -393,7 +413,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 24,
+   "execution_count": 34,
    "metadata": {},
```
mat10d, line 417: Looks good overall. Just one thing--I think we should add some further things to the config file:

        z_interval=4,
        channel_order_flip=True,
        
        These are both part of the preprocess.smk. You want to ask the user if they have multiple z stacks, in order to flatten the metadata. And also if their images are in a standard order (with Dapi first), then flip would be false. Otherwise, if they are reversed, it would be true. Otherwise, good to go. Good prompt engineering.

mat10d (changes requested):
Need to add a few more values to the config file and have these be read in by preprocess.smk, if this is to scale to others' analyses.

roshankern:
@mat10d Thanks for pointing this out! I've added the relevant code for these params. Please make sure that my description of these params is correct in the configuration notebook.

roshankern:
Started these changes from the wrong base branch. Will reopen in another PR to avoid merge conflicts.

PR #29, opened 2024-12-16
Preprocessing fixes

mat10d:
Seeking to make the following changes. When the boxes are ticked, they are ready for review:
- [x] Better information in the 0.ipynb. There was confusion over: 1) how to use the LLM--made it clear that it is to generate a good regex and 2) whether SBS_SAMPLES_DF_FP/PHENOTYPE_SAMPLES_DF_FP were generated or user made (these are generated by the script). Made these small updates. _One thing to note is that it seems like the example_analysis notebook for this step is missing the regex that we used. I think we want to make it clear that one can download example data and run the analysis from this in the example_data folder (this is powerful because it shows a user that the pipeline) and leave it black in the analysis folder). @roshankern can you check if this was overwritten? Now seeing that you did the same for all the other steps. Need to revert!
- [x] Head README changes: 1) there were some typos on the readme. 2) we want to provide information on how to run just sbs or phenotype analysis. I added the info you gave me, but @roshankern it may be smart to test if this is actually the case (missing files -> missing sample tsv -> downstream rules don't run; --until flags are working properly)
- [x] Other aspects of the analysis notebooks don't make sense. A user wouldn't test empty vs real image tiles, so it's unnecessary to have both of these options in the analysis subfolder. It would be helpful if @roshankern did a deeper dive into the current implementations of analysis/ and made sure that it was logical to the user, and restored example_analysis/ to having all the file paths to ensuring it runs
- [ ] Different preprocessing inputs. If we stick with just tile based files, people could still just generate one image per tile per channel. We will need to handle this. There is also the chance that there is more than one round of imaging. In both cases, we just need to add a way of flattening multiple tiffs into 1. This is ongoing...


roshankern:
> @roshankern can you check if this was overwritten? Now seeing that you did the same for all the other steps. Need to revert!

I was overwritten somehow. Fixed in https://github.com/cheeseman-lab/brieflow/pull/29/commits/8010c6aca96a550978b1cbd1cfac2c92ad4d6665

roshankern:
> It would be helpful if @roshankern did a deeper dive into the current implementations of analysis/ and made sure that it was logical to the user, and restored example_analysis/ to having all the file paths to ensuring it runs

Made changes to 2 notebooks in https://github.com/cheeseman-lab/brieflow/pull/29/commits/297f108bc7b30d1708ecded4211047648f002df3, let me know if anything is still unclear.

roshankern:
> Different preprocessing inputs. If we stick with just tile based files, people could still just generate one image per tile per channel. We will need to handle this. There is also the chance that there is more than one round of imaging. In both cases, we just need to add a way of flattening multiple tiffs into 1. This is ongoing...

As per our Slack discussion, we are leaving this for another PR. This PR is ready for review @mat10d !

README.md
roshankern, line 1: Thank you for fixing these typos!

roshankern (approved):
LGTM!

README.md
```diff
@@ -117,16 +117,20 @@ sh 1.run_preprocessing.sh
 ```
 **Slurm**:
 ```sh
-sh 1.run_preprocessing_slurm.sh
+sbatch 1.run_preprocessing_slurm.sh
 ```
 
+***Note**: For testing purposes, users may only have generated sbs or phenotype images.
+If this is the case, and one of the `SBS_SAMPLES_DF_FP`/`PHENOTYPE_SAMPLES_DF_FP` are empty, then this will impede any further analysis of the missing files.
```
mat10d, line 124: Is this accurate? I noticed you changed the if statements for sbs and phenotype in the snakefile, but will things also be ok at preprocessing?

Only comment I have before merging is to ensure that it is very clear how the analysis runs if 1) you are missing one data type or 2) you want to just one 1/2 analyses

mat10d (commented):
Just one comment

README.md
```diff
@@ -117,16 +117,21 @@ sh 1.run_preprocessing.sh
 ```
 **Slurm**:
 ```sh
-sh 1.run_preprocessing_slurm.sh
+sbatch 1.run_preprocessing_slurm.sh
 ```
 
+***Note**: For testing purposes, users may only have generated sbs or phenotype images.
+If either of the dataframes defining the file paths and metadata for sbs/phenotype samples (located at `SBS_SAMPLES_DF_FP`/`PHENOTYPE_SAMPLES_DF_FP`) are empty, then no preprocessing steps will be run for the respective missing data type.
+Use an empty dataframe for the sbs/phenotype samples to skip preprocessing for a particular data type.
```
mat10d, line 125: Ok, good to know. Can you also restrict this using an until flag?

README.md
```diff
@@ -137,18 +142,22 @@ sh 4.run_sbs_phenotype_processes.sh
...
 **Slurm**:
 ```sh
-sh 4.run_sbs_phenotype_processes_slurm.sh
+sbatch 4.run_sbs_phenotype_processes_slurm.sh
 ```
 
 ***Note**: Use `brieflow_configuration` Conda environment for each configuration notebook.
 
+***Note**: Many users will want to only run SBS or phenotype processing, independently.
+By varying the tags in the .sh files (`--until all_sbs_process` or `--until all_phenotype_process`), the analysis will only run only the analysis of interest.
```
mat10d, line 151: I would align this explanation to the prior:
1) if there are no files, it won't run
2) if you want to manually restrict it from running, use the until flag

README.md
roshankern: Not as of now. We could create two additional rules in `workflow/rules/preprocess.smk` for the following targets:
1) PREPROCESS_TARGETS_SBS
2) PREPROCESS_TARGETS_PHENOTYPE

Then one should be able to restrict preprocessing using an until flag. Should I add these two rules?

README.md
roshankern: Done in https://github.com/cheeseman-lab/brieflow/pull/29/commits/4cf230dd1bb9f2324e714f0950c08a582f1ed673

README.md
mat10d: We can think about that if it makes sense down the road. If we can't use until flags for this, that's fine--don't add this into the readme then

PR #31, opened 2024-12-18
Preprocessing multichannel

mat10d:
After thinking about it harder, adding a multichannel file reader is not really in the area of "new functionality", because doing so would have to rewrite the nd2_to_tiff and in terms of scope, does little to actually improve functionality--it is moreso something that we need. Therefore, I think we should focus on migrating a segmentation method as a way of documenting this to others. 

Anyway--this is ready for review and we should ping @pupasam once we're done.


roshankern (approved):
I made some changes for you to review/test. If these work then LGTM!

mat10d:
Great. I checked with both datasets and it worked.

PR #44, opened 2025-02-10
Well merge

mat10d:
# Description

Thank you for your contribution to Brieflow!
Please _succinctly_ summarize your proposed change.
What motivated you to make this change?

Changes from well based processing, specifically missing any well specific:
- Snakefiles
- Rules
- Targets

These will all stay on well branch, but other changes can be merged into main.

Note: also deleted the analysis/ folder


example_analysis/1.run_preprocessing_slurm.sh
```diff
@@ -22,3 +22,8 @@ snakemake --executor slurm --use-conda \
     --configfile "config/config.yml" \
     --latency-wait 60 \
     --until all_preprocess
+
+# End timing and calculate duration
+end_time=$(date +%s)
+duration=$((end_time - start_time))
+echo "Total runtime: $((duration / 3600))h $(((duration % 3600) / 60))m $((duration % 60))s" >> slurm/slurm_output/main/preprocessing-$SLURM_JOB_ID.out
```
roshankern, line 29: Would be good to follow line at EOF conventions for these sh files.

roshankern, line 29: Really like this change to time the steps, will be good for benchmarking. Long term it might be worth looking at [snakemake benchmarking](https://snakemake.readthedocs.io/en/stable/tutorial/additional_features.html#benchmarking) if this information is very important.

example_analysis/slurm/config.yaml
```diff
@@ -80,22 +87,38 @@ set-resources:
...
+        mem_mb: 10000
+    eval_merge:
+        mem_mb: 10000
+    clean_merge:
+        mem_mb: 10000
+    deduplicate_merge:
+        mem_mb: 10000
+        cpus_per_task: 4
+    final_merge:
+        mem_mb: 500000
```
roshankern, line 124:
```suggestion
        mem_mb: 500000

```

roshankern: Thanks for figuring out these memory requirements 😃

workflow/Snakefile
```diff
@@ -19,7 +19,7 @@ SBS_SAMPLE_FPS = list(sbs_samples_df["sample_fp"].unique())
 if len(SBS_SAMPLE_FPS) > 0:
     SBS_WELLS = list(sbs_samples_df["well"].unique())
     SBS_TILES = list(sbs_samples_df["tile"].unique())
-    SBS_CYCLES = list(sbs_samples_df["cycle"].unique())
+    SBS_CYCLES = sorted(list(sbs_samples_df["cycle"].unique()))
```
roshankern, line 22: What was the logic behind this?

We should be sorting by all metadata in `create_samples_df`. Maybe there is a bug in this function?

workflow/lib/merge/eval_merge.py
```diff
@@ -133,3 +133,29 @@ def plot_channel_histogram(df_before, df_after, channel_min_cutoff=0):
...
+    # Plot scatter with either fixed color or hue-based coloring
+    if color is not None:
+        sns.scatterplot(data=df, x="i_0", y="j_0", color=color, alpha=0.5)
+    else:
+        sns.scatterplot(data=df, x="i_0", y="j_0", hue=hue, alpha=0.5)
+
+    plt.title(title)
+    plt.xlabel("i_0")
+    plt.ylabel("j_0")
+    return fig
```
roshankern, line 161:
```suggestion
    return fig

```

workflow/lib/merge/eval_merge.py
```diff
@@ -133,3 +133,29 @@ def plot_channel_histogram(df_before, df_after, channel_min_cutoff=0):
     plt.ylabel("Count")
     plt.legend()
     return fig
+
+
+def plot_cell_positions(df, title, color=None, hue="channels_min"):
+    """Generates a scatter plot of cell positions in the i_0, j_0 coordinate space.
+
+    Args:
+        df: DataFrame containing cell position data with i_0, j_0 columns.
```
roshankern, line 142: Could you make the parameter name more specific than df?

workflow/lib/phenotype_process/align_channels.py
```diff
@@ -0,0 +1,75 @@
...
+    source,
+    riders=[],
+    upsample_factor=2,
+    window=2,
+    remove_channel=False,
+):
+    """Rigid alignment of phenotype channels based on target and source channels.
+
+    Args:
+        data (np.ndarray): The input data containing the channels with dimensions
```
roshankern, line 25: would be good to have a more specific param name if possible

workflow/lib/preprocess/preprocess.py
```diff
@@ -83,6 +85,93 @@ def extract_tile_metadata(
     return df
 
 
+def extract_well_metadata(
```
roshankern, line 88: I thought the plan was to not add well-specific functions to main, at least those that we would not expect a broad audience to use in the future. As part of our attempt to keep the main codebase clean and concise while having useful functionality.

Is my understanding of this off?

workflow/lib/preprocess/preprocess.py
```diff
@@ -153,3 +242,119 @@ def nd2_to_tiff(
         print(f"Final dimensions (CYX): {result.shape}")
 
     return result.astype(np.uint16)
+
+
+def nd2_to_tiff_well(
```
roshankern, line 247: Same comment as above.

workflow/lib/shared/target_utils.py
```diff
@@ -88,3 +88,55 @@ def output_to_input(output_path, wildcard_values, wildcards):
...
+        wells (list): List of well identifiers.
+        tiles (list, optional): List of tile identifiers. Defaults to None.
+
+    Returns:
+        list: List of fully resolved target paths.
+    """
+    targets = []
+    for output_template in outputs:
+        for well in wells:
+            for combo in valid_combinations:
```
roshankern, line 128: A little cleaner to use [itertools.product](https://stackoverflow.com/questions/22473053/itertools-equivalent-of-nested-loop-for-x-in-xs-for-y-in-ys) for iterating over multiple things. Maybe not worth a change here but worth keeping in mind for future code.

workflow/rules/merge_process.smk
```diff
@@ -7,13 +7,17 @@ rule fast_alignment:
...
+        lambda wildcards: output_to_input(
+            PREPROCESS_OUTPUTS["combine_metadata_phenotype"],
+            {} if config["merge_process"]["ph_metadata_channel"] is None else {"channel": config["merge_process"]["ph_metadata_channel"]},
+            wildcards,
+        ),
         lambda wildcards: output_to_input(
             PREPROCESS_OUTPUTS["combine_metadata_sbs"],
-            {"cycle": config["merge_process"]["sbs_metdata_cycle"]},
+            {"cycle": config["merge_process"]["sbs_metadata_cycle"], 
+             **({} if config["merge_process"]["sbs_metadata_channel"] is None else {"channel": config["merge_process"]["sbs_metadata_channel"]})},
```
roshankern, line 18: Are these changes specific to well-based processing? I'm having some trouble understanding what's happening here... maybe we can discuss in person?

If we don't expect this approach to be used by most people processing OPS data than better to leave it out and keep things simple imo.

workflow/rules/phenotype_process.smk
```diff
@@ -13,13 +13,29 @@ rule apply_ic_field_phenotype:
...
+    input:
+        PHENOTYPE_PROCESS_OUTPUTS["apply_ic_field_phenotype"]
+    output:
+        PHENOTYPE_PROCESS_OUTPUTS_MAPPED["align_phenotype"]
+    params:
+        align=config["phenotype_process"]["align"],
+        target=lambda wildcards: config["phenotype_process"]["target"] if config["phenotype_process"]["align"] else None,
+        source=lambda wildcards: config["phenotype_process"]["source"] if config["phenotype_process"]["align"] else None,
+        riders=lambda wildcards: config["phenotype_process"]["riders"] if config["phenotype_process"]["align"] else None,
+        remove_channel=lambda wildcards: config["phenotype_process"]["remove_channel"] if config["phenotype_process"]["align"] else None,
```
roshankern, line 29: This is sick, good idea to dynamically include params here. But, do we need `lambda wildcards:` here? It seems like we shouldn't need it if we aren't referencing the `wildcards` var.

workflow/scripts/merge_process/deduplicate_merge.py
roshankern, line 1: Am I missing something or are these changes a mistake? Looks like we should still be starting from `input[0]`

image

workflow/scripts/merge_process/final_merge.py
roshankern, line 1: Same comment as above

image

workflow/scripts/phenotype_process/align_phenotype.py
```diff
@@ -0,0 +1,23 @@
+from tifffile import imread, imwrite
+
+from lib.phenotype_process.align_channels import align_phenotype_channels
```
roshankern, line 3: Thanks for following import conventions!

workflow/scripts/preprocess/extract_well_metadata.py
roshankern, line 1: Is this only useful for well base processing? Do we need this in main branch?

workflow/scripts/preprocess/nd2_to_tiff_well.py
roshankern, line 1: Same as above

workflow/scripts/sbs_process/apply_ic_field_sbs.py
```diff
@@ -1,15 +1,22 @@
 from tifffile import imread, imwrite
 
-from lib.shared.illumination_correction import apply_ic_field
+from lib.shared.illumination_correction import apply_ic_field, combine_ic_images
 
 # load aligned image data
 aligned_image_data = imread(snakemake.input[0])
 aligned_image_data_segmentation_cycle = aligned_image_data[
-    snakemake.params.segmentation_cycle_index - 1
+    snakemake.params.segmentation_cycle_index
```
roshankern, line 8: Good catch

roshankern (changes requested):
Requesting some changes before merge here. Feel free to ask me about this in person as well.

example_analysis/1.run_preprocessing_slurm.sh
mat10d: Good to think about for the future

workflow/scripts/merge_process/final_merge.py
mat10d: Cleaned this up in the rules file

workflow/scripts/merge_process/deduplicate_merge.py
mat10d: Cleaned this up in the rules file

PR #83, opened 2025-04-01
Aggregate/cluster overhaul

roshankern:
# Description

Overhaul of aggregate and cluster processes to use TVN. Cluster process also has new (although based on old) evaluation metrics.

- [ ] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

- [] My code follows the style guidelines of this project.
- [] I have checked linting and formatting with `ruff check` and `ruff format`.
- [] I have performed a self-review of my own code.
- [] I have commented my code, particularly in hard-to-understand areas.
- [] I have made corresponding changes to the documentation.
- [] My changes generate no new warnings.
- [] I have deleted all non-relevant text in this pull request template.


workflow/scripts/aggregate/align.py
mat10d, line 1: If this is the data bottleneck step, I would perhaps maintain the gc.collect calls but make them optional via a verbose flag. This could be exposed to the user, maybe you should just keep it in the snakemake?

workflow/scripts/aggregate/eval_aggregate.py
mat10d, line 1: Yeah, worried again about the outputs of this and what it tells us...

mat10d (approved):
Overall, quite good with some minor changes to make.

brieflow_main_env.yml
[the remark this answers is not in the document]
roshankern: We do, good catch! Added in https://github.com/cheeseman-lab/brieflow/pull/83/commits/1807d617884c536bc35fa9c7684d275bbb2f2c1e

workflow/scripts/aggregate/eval_aggregate.py
roshankern: Resolved

workflow/Snakefile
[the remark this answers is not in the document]
roshankern: Yep will do this before merging

workflow/lib/cluster/benchmark_clusters.py
mat10d, line 1: Need better annotations here!

workflow/lib/cluster/benchmark_clusters.py
```diff
@@ -1,366 +1,396 @@
...
-    return cluster_summary
+    return pair_recall_global["recall"].mean()
 
 
-def analyze_differential_features(
-    cluster_gene_table,
-    feature_df,
-    n_top=5,
-    exclude_cols=["gene_symbol_0", "cell_number"],
+def calculate_group_enrichment(
```
mat10d, line 160: Missing docstrings for this function

workflow/lib/cluster/benchmark_clusters.py
```diff
@@ -1,366 +1,396 @@
...
+            break
+
+    cluster_enrichment = (
+        pd.DataFrame(enriched_rows).sort_values("cluster").reset_index(drop=True)
+    )
+
+    return cluster_enrichment["num_enriched_groups"].mean()
+
+
+def perform_resolution_thresholding(
```
mat10d, line 223: Missing docstrings

mat10d (approved):
Overall, looks pretty good. The annotation for functions in aggregate were awesome, we want to maintain that in clustering. This is especially true for the benchmarking--what statistical methodology you were using is pretty critical to tell the user

I also may suggest that for the uniprot mapping, you have a way of mapping not only on the gene name but also on the uniprot id if this is in the original table. This is something we will want to integrate early on, and so would be good to have it already set up

workflow/lib/cluster/benchmark_clusters.py
roshankern: Added in https://github.com/cheeseman-lab/brieflow/pull/83/commits/a59120d2a0a70a164a19342f6c194be3a0c52122

workflow/lib/cluster/benchmark_clusters.py
roshankern: Added in https://github.com/cheeseman-lab/brieflow/pull/83/commits/a59120d2a0a70a164a19342f6c194be3a0c52122

workflow/scripts/aggregate/align.py
roshankern: Most of these were unnecessary so I've removed all but 2 and don't think it makes sense to provide this as an exposed param now.

workflow/lib/cluster/benchmark_clusters.py
roshankern: Yep, I've added specifics on the statistical methodologies in https://github.com/cheeseman-lab/brieflow/pull/83/commits/e179c077557a08c85d64e49dfede903e13fee147

roshankern:
> Overall, looks pretty good. The annotation for functions in aggregate were awesome, we want to maintain that in clustering. This is especially true for the benchmarking--what statistical methodology you were using is pretty critical to tell the user

Good idea -- done!

> I also may suggest that for the uniprot mapping, you have a way of mapping not only on the gene name but also on the uniprot id if this is in the original table. This is something we will want to integrate early on, and so would be good to have it already set up

The uniprot ID is not in the original table :( so let's save this for later

PR #146, opened 2025-07-24
color code twin y ax in read mapping qc

rtwalton:
color code twin y axes in read mapping QC plot for intuitive interpretation of "read mapping (%) " vs "reads mapped (count)"


mat10d:
Thanks for this contribution @rtwalton. Please run `ruff format` from within `brieflow` after activating your brieflow environment to reformat appropriately, and repush, so we can accept!

mat10d:
Handled in #161 with ruff checks

PR #160, opened 2025-10-27
Add support for z-split TIFF input files

mat10d:
## Summary
- Add support for z-split TIFF input files in preprocessing
- Update `get_sample_fps()` to handle z-dimension filtering and return sorted z-plane lists
- Enhance `convert_nd2_to_array_well()` and `convert_nd2_to_array_tile()` to support z-stacking
- Improve handling of z-planes when channel order is specified

Ready for review.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "BrieFlow"
-version = "1.2.0"
+version = "1.3.2"
 description = "Package for processing data from optical poooled screens."
 readme = "README.md"
 license = { file = "LICENSE" }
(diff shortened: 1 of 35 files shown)


workflow/lib/preprocess/preprocess.py
```diff
@@ -729,31 +731,67 @@ def convert_nd2_to_array_well(
...
         from tifffile import imread
     except ImportError:
         raise ImportError("tifffile package required for TIFF support")
 
+    # Handle deprecated z_stack parameter
+    if "z_stack" in kwargs and kwargs["z_stack"] is True and n_z_planes is None:
+        if verbose:
+            print(
+                "Warning: z_stack=True is deprecated. Use n_z_planes parameter instead."
+            )
```
roshankern: It is odd to me to add deprecated code in a new PR. Is there a reason to simply not add this deprecated param and simply force users to use the newest code?

roshankern (approved):
Some small comments. Once these are fixed/addressed, LGTM!

workflow/lib/preprocess/preprocess.py
mat10d: Removed, good point, was an artifact.

PR #193, opened 2026-02-09
Adding Imaris File Reading to Brieflow supported filetypes.

joandrews33:
# Description

Adding an Imaris File reader to extend supported file types. Reads in metadata bundled with each Imaris image file, and converts images to tiff files.

I also added support in the workflow/scripts/preprocessing/extract_metadata.py to allow the ability to use both bundled metadata and a separate metadata file together when extracting metadata (because Imaris has per image and additional per job metadata files), but I wasn't able to get this to integrate with snakemake without making deeper structural changes to the code base. I commented out the new code, and reverted the behavior to legacy out of worry for breaking support for workflows that need a separated metadata file, but if it works with those workflows, it may be useful new functionality. 

Enhancement (adds functionality)

# Checklist

- [Not updated. I am a few updates behind main.] I have updated the `pyproject.toml` to reflect the change as designated by [semantic versioning](https://semver.org/).
- [ N/A ?] I have made corresponding changes to the documentation.
- [x] I have deleted all non-relevant text in this pull request template.


joandrews33:
Ready for review. Changes to scripts/preprocessing/extract_metadata.py can be safely rejected, as my pipeline doesn't need them as is, but they would be necessary if I wanted to incorporate automatic pixel size determination, which requires cross-referencing two different metadata files.

mat10d:
Picking this back up. This branch had fallen behind `main`, which is why the `Test Analysis` gate never ran on it — that workflow only runs on PRs to `main` that are up to date.

Rather than push to your branch, I've opened #264, which is `main` merged with `joandrews33:ims-reader` plus a couple of small follow-up commits. All of your commits are in it with your authorship, so merging #264 will mark this PR merged too. Nothing here has been changed.

The merge was clean — `preprocess.py` hadn't moved on `main` since your branch point. My follow-ups were: lifting the nested HDF5 helpers to module level, making the missing-attribute cases return `None` instead of raising on `float(None)`, `Path`-safe `F<number>` tile parsing with a fallback, keeping `FieldID` in its own `field_id` column so `well` stays the configured well that `combine_metadata` filters on, passing `cycle`/`round` through, forwarding `**kwargs` in `convert_to_array`, and a docs line. Full list and reasoning in the #264 description.

The metadata-file-plus-bundled-metadata idea you described in the PR body is still open — it needs a rule-level change so `extract_metadata` can take `samples` and `metadata` inputs together, so it's worth its own PR. Details are in #264 too.

mat10d:
Update: the landing PR moved from #264 (main) to #265 (zarr3), because zarr3 merges into main next week. Your commits are carried unchanged with your authorship; merging #265 will mark this PR as merged.

PR #206, opened 2026-03-27
fix dtype to be spec compliant, fix gaussian scaling for saving images

mat10d:
Fixes minor spec compliant issues that merited more digging.

### DIFF
--- a/workflow/lib/shared/io.py
+++ b/workflow/lib/shared/io.py
@@ -242,7 +242,7 @@ def write_image_omezarr(
         axes=axes,
         coordinate_transformations=coordinate_transformations,
         scaler=Scaler(
-            method="nearest",
+            method="nearest" if is_label else "gaussian",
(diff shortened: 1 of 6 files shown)


workflow/lib/shared/io.py
```diff
@@ -242,7 +242,7 @@ def write_image_omezarr(
         axes=axes,
         coordinate_transformations=coordinate_transformations,
         scaler=Scaler(
-            method="nearest",
+            method="nearest" if is_label else "gaussian",
```
exxpyy, line 245: Note that in schema-updates branch I'm working on there's a downsamplingMethod field written to multiscales metadata that we should match with this

exxpyy (commented):
Looks good, but we need to make sure the downsamplingMethod metadata field in schema-updates gets updated to match

mat10d:
Changes merged into zarr3-transition via schema-updates.

Issue #198, opened 2026-02-17
[BUG] align rule crashes with ValueError: array must not contain infs or NaNs in tvn_on_controls


mat10d:
## Summary

The `align` rule crashes with `ValueError: array must not contain infs or NaNs` when `tvn_on_controls()` encounters a singular or near-singular covariance matrix, typically from cell classes with very few control cells (e.g., Mitotic).

## Environment

- **Script**: `brieflow/workflow/scripts/aggregate/align.py`
- **Library**: `brieflow/workflow/lib/aggregate/align.py:180`
- **Rule**: `align` in `brieflow/workflow/rules/aggregate.smk`
- **Screen**: shasta-analysis (Mitotic cell class)

## Error

```
Traceback (most recent call last):
  File ".snakemake/scripts/tmpq69oyt70.align.py", line 122, in <module>
    features = tvn_on_controls(
  File "brieflow/workflow/lib/aggregate/align.py", line 180, in tvn_on_controls
    embeddings[batch_ind], linalg.fractional_matrix_power(source_cov, -0.5)
  ...
  File "numpy/lib/_function_base_impl.py", line 649, in asarray_chkfinite
    raise ValueError(
ValueError: array must not contain infs or NaNs
```

## Root Cause

`tvn_on_controls()` at `align.py:180` computes `linalg.fractional_matrix_power(source_cov, -0.5)` which produces infs/NaNs when the source covariance matrix is singular or near-singular. This is then caught by `numpy.asarray_chkfinite`.

This occurs when a cell class (e.g., Mitotic) has very few control cells, resulting in a degenerate covariance matrix that cannot be inverted.

Both Mitotic channel combos failed identically in the shasta screen (with and without KI67 marker).

## Proposed Solution

Add a check for singular/near-singular covariance matrices before computing the fractional matrix power. Options:
1. Check the condition number of `source_cov` and skip the batch (with a warning) if it's too high
2. Use a regularized pseudoinverse instead of `fractional_matrix_power`
3. Add a small ridge (epsilon * I) to `source_cov` before inversion to ensure numerical stability

Issue #196, opened 2026-02-12
[BUG] KeyError when `perturbation_id_col` is None


akcd1:
## Summary
Script `generate_feature_table.py` crashes with `KeyError: None` when the optional config parameter `perturbation_id_col` is set to `None`.

## Environment
- **Script**: `brieflow/workflow/scripts/aggregate/generate_feature_table.py`
- **Rule**: `generate_feature_table` in `brieflow/workflow/rules/aggregate.smk`

## Expected Behavior
According to the documentation in `8.configure_aggregate_params.ipynb`:
> `PERTURBATION_ID_COL`: Name of column that identifies unique perturbations. **Only needed if you want your controls to have different perturbation names**, ex `cell_barcode_0`. **Otherwise, can leave this as `None`**.

When `perturbation_id_col` is `None`, the script should use `perturbation_name_col` (gene name) as the construct identifier.

## Actual Behavior
The script crashes with:
```
KeyError: None

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "generate_feature_table.py", line 129, in <module>
    for construct_id in metadata[pert_id_col].unique():
```

## Root Cause
The script attempts to access `metadata[None]` when `pert_id_col` is `None`, treating it as a column name instead of recognizing it as a missing optional parameter.

**Affected lines**: 129, 130, 162, 177, 245, 248, 296, 299, 312

## Fix Applied
1. Added `construct_id_col = pert_id_col if pert_id_col is not None else pert_col` at line 23
2. Replaced column access with `construct_id_col` where appropriate
3. Conditionally structured output tables based on whether `pert_id_col` exists

## Test Case
**Config with bug**:
```yaml
perturbation_name_col: gene_symbol_0
perturbation_id_col: null  # Should be valid but causes crash
```

**Config without bug** (workaround before fix):
```yaml
perturbation_name_col: gene_symbol_0
perturbation_id_col: cell_barcode_0  # Must specify even if not needed
```

Issue #195, opened 2026-02-12
[BUG] `split_cell_data` crashes when cell class is absent from a well


akcd1:
## Description

The aggregate filter script (`brieflow/workflow/scripts/aggregate/filter.py`) crashes with a `ValueError` when processing wells that contain no cells of a specific cell class.

## Environment

- BrieFlow version: 1.4.9
- Python version: 3.11

## Error Message

```
ValueError: Found array with 0 sample(s) (shape=(0, 24)) while a minimum of 1 is required by LocalOutlierFactor.
```

Full traceback:
```
Traceback (most recent call last):
  File ".snakemake/scripts/tmp44x3y8oq.filter.py", line 41, in <module>
    metadata, features = intensity_filter(
                         ^^^^^^^^^^^^^^^^^
  File "briefflow/workflow/lib/aggregate/filter.py", line 227, in intensity_filter
    ).fit_predict(features[intensity_cols])
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../sklearn/neighbors/_lof.py", line 258, in fit_predict
    return self.fit(X)._predict()
           ^^^^^^^^^^^
  File ".../sklearn/base.py", line 1336, in wrapper
    return fit_method(estimator, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../sklearn/neighbors/_lof.py", line 281, in fit
    self._fit(X)
  File ".../sklearn/neighbors/_base.py", line 519, in _fit
    X = validate_data(
        ^^^^^^^^^^^^^^
  File ".../sklearn/utils/validation.py", line 2902, in validate_data
    out = check_array(X, input_name="X", **check_params)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../sklearn/utils/validation.py", line 1097, in check_array
    raise ValueError(
ValueError: Found array with 0 sample(s) (shape=(0, 24)) while a minimum of 1 is required by LocalOutlierFactor.
```

## How to Reproduce

1. Configure cell classification with multiple cell classes
2. Have a well where one or more cell classes are naturally absent
3. Run the aggregate pipeline
4. The pipeline generates empty parquet files for absent cell classes during `split_datasets`
5. The `filter` rule attempts to process these empty files and crashes at the `intensity_filter` step

## Root Cause

[...]

PR #208, opened 2026-03-30
Fix aggregate edge cases (#195, #196, #198)

mat10d:
# Description

Thank you for your contribution to Brieflow!
Please _succinctly_ summarize your proposed change.
What motivated you to make this change?

Fixes three aggregate module crashes that occur with sparse or edge-case data:

- #195: intensity_filter crashes with LocalOutlierFactor on 0 samples when a cell class is absent from a well. Added early return on empty data and write_empty_and_exit guards after each filter step.
- #196: generate_feature_table.py and prepare_bootstrap_data.py crash with KeyError: None when perturbation_id_col is None. Falls back to perturbation_name_col.
- #198: tvn_on_controls crashes with ValueError: array must not contain infs or NaNs when too few control cells. Skips TVN with 0 controls, skips per-batch CORAL when controls < features or covariance is singular.

Validated on real data from the viso screen (lourido lab) — empty cell classes (Uninfected wells with 0 cells) and wells with 0 controls.

Files changed

- workflow/lib/aggregate/filter.py — early return on empty data
- workflow/lib/aggregate/align.py — 0-control guard + sparse batch CORAL skip
- workflow/scripts/aggregate/filter.py — write_empty_and_exit after each filter step
- workflow/scripts/aggregate/generate_feature_table.py — pert_id_col fallback
- workflow/scripts/aggregate/prepare_bootstrap_data.py — pert_id_col fallback

Please also link to any relevant issues that your code is associated with.

- [x] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

- [x] I have updated the `pyproject.toml` to reflect the change as designated by [semantic versioning](https://semver.org/).
- [x] I have made corresponding changes to the documentation.
- [x] I have deleted all non-relevant text in this pull request template.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "BrieFlow"
-version = "1.4.10"
+version = "1.4.11"
 description = "Package for processing data from optical poooled screens."
 readme = "README.md"
 license = { file = "LICENSE" }
--- a/workflow/lib/aggregate/align.py
+++ b/workflow/lib/aggregate/align.py
@@ -157,8 +157,12 @@ def tvn_on_controls(
     Returns:
         np.ndarray: The normalized embeddings.
     """
-    embeddings = centerscale_on_controls(embeddings, metadata, pert_col, control_key)
     ctrl_ind = metadata[pert_col].str.startswith(control_key).to_list()
+    n_controls = sum(ctrl_ind)
+    if n_controls == 0:
+        print("Warning: no control cells found, skipping TVN normalization")
+        return embeddings
+    embeddings = centerscale_on_controls(embeddings, metadata, pert_col, control_key)
     embeddings = PCA().fit(embeddings[ctrl_ind]).transform(embeddings)
     embeddings = centerscale_on_controls(
         embeddings, metadata, pert_col, control_key, batch_col
--- a/workflow/lib/aggregate/align.py
+++ b/workflow/lib/aggregate/align.py
@@ -141,6 +141,7 @@ def tvn_on_controls(
     pert_col: str,
     control_key: str,
     batch_col: str | None = None,
+    control_col: str | None = None,
 ) -> np.ndarray:
     """Apply TVN (Typical Variation Normalization) to the data based on the control perturbation units.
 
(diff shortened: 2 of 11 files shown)


akcd1 (approved):
The three crash scenarios addressed here have good fixes. LGTM.

Two failure modes in the same sparse-data scenario aren't addressed by this PR and will still crash the pipeline after merge:

---

### 1. `workflow/lib/aggregate/filter.py` — `ValueError` when KNN imputation runs on integer-typed feature columns

`KNNImputer.fit_transform()` always returns a `float64` numpy array. If any feature column with remaining NAs is backed by an integer dtype (e.g. count-based CellProfiler measurements), writing the imputed values back into that column raises:
```
ValueError: Cannot convert non-finite values (NA or inf) to integer
```

The fix is to cast any integer column that still has NAs to `float64` before the imputer runs:
```python
for col in remaining_cols_with_na:
    if pd.api.types.is_integer_dtype(features[col]):
        features[col] = features[col].astype("float64")
```

This needs to go in `missing_values_filter()` in `lib/aggregate/filter.py`, just before the `has_na_mask` line. Since this PR is already touching the empty-data path in the same function, it would be a natural addition here.

---

### 2. `workflow/scripts/aggregate/align.py`, `eval_aggregate.py`, `generate_feature_table.py` — `ArrowInvalid` schema conflict when mixing empty and non-empty parquet files

The per-well filter step (which this PR correctly guards) can still produce empty parquet files in some cases — for example when all cells are removed by `query_filter` or `perturbation_filter` rather than at the initial row-count check. PyArrow writes empty parquets with null-typed columns because it can't infer schema from zero rows. When `ds.dataset()` later opens a directory containing both real (typed) and empty (null-typed) files, it raises:
```
pyarrow.lib.ArrowInvalid: Schema at index 0 was different: ...
```

This affects any downstream script that opens a collection of per-well parquets via `ds.dataset()`. The fix is to collect schemas from non-empty files first and pass the unified schema explicitly:
```python
all_schemas = [pq.read_table(p).schema for p in paths if pq.read_metadata(p).num_rows > 0]
if all_schemas:
    cell_dataset = ds.dataset(paths, format="parquet", schema=all_schemas[0])
else:
    cell_dataset = ds.dataset(paths, format="parquet")
```

This pattern needs to be applied in `align.py`, `eval_aggregate.py`, and `generate_feature_table.py` wherever `ds.dataset()` is called on a list of per-well paths.

mat10d:
Recent commits add fixes to those issues, along with #215 and #214 @akcd1

akcd1:
LGTM

PR #256, opened 2026-08-27
feat(aggregate): split or group by any label column, with the bootstrap following the grouping

mat10d:
Screens increasingly carry per-well experimental variables that are not perturbations — zargun assigns one of six treatments per well, potc varies confluency. The analysis they call for is a representation per `perturbation x treatment`, matching the reference analysis, whose core operation is

```python
.groupby(["gene_name", "well", "treatment", "name"]).agg(meanval=..., sdval=..., semval=...)
```

## Concept

Splitting and grouping are two independent operations over *any* label column, and the column's provenance is irrelevant to either:

| | effect | embedding space |
|---|---|---|
| **split** | one dataset per value, full downstream pipeline each | separate per value |
| **group** | one dataset, extra key in the aggregation groupby | shared |

Labels come from a plate map joined on `(plate, well)` — treatment, confluency, timepoint — or from a classifier, as `cell_class` already does. Choosing split vs group is a per-screen analysis decision, not a property of the label. Splitting infected/uninfected asks what the landscape looks like within each population; grouping by it asks how a perturbation differs between them.

## Config (all default to current behaviour)

```
aggregate.well_annotations_fp       plate map TSV (plate, well, + annotation cols)   None
aggregate.split_col                 column to subset on                              "class"
aggregate.group_cols                extra aggregation keys                           []
aggregate.bootstrap_control_scope   "pooled" | "within_group"                        "pooled"
```

A pure drug screen needs only the join: set `perturbation_name_col: treatment`, `control_key: Ethanol`, and the existing single-key groupby is already correct.

## Bootstrap

The bootstrap does not consume `lib/aggregate/aggregate.py` — its points come from `generate_feature_table.py`. Making the null follow the grouping therefore means grouping the construct and gene tables, done with a **composite key** (`MYC=Corticosterone`) so no output paths or wildcards change. The two-stage structure is preserved: `bootstrap_construct` runs per construct x group, and `bootstrap_gene` globs `{gene}__*`, which now resolves per treatment.

`bootstrap_control_scope: within_group` is not cosmetic. `run_construct_bootstrap` treats the control label as opaque, but `_bootstrap_inner_numba` picks its seed row uniformly across *all* control rows — so with composite labels a construct would silently draw its null from a random treatment. Measured on a synthetic screen with controls at 10.0 under one treatment and 0.0 under the other:

```
pooled        null mean 5.18   sd 4.99    48.1% of sims drew the wrong treatment
within_group  null mean 9.98   sd 0.163    0.0%

p-values, four genuine hits:
  pooled        0.9975  0.9940  0.9945  0.9940
  within_group  0.0105  0.0000  0.0065  0.0000
```

## Also fixed

`bootstrap_gene.py` hardcoded `"gene_symbol_0"` in three places, so a drug-only screen configured with `perturbation_name_col: treatment` raised `KeyError` — pre-existing and independent of this feature.

`control_mask` in lib replaces `.str.contains(control_key)` at the sites that see composite keys. Matching the whole key lets a group value satisfy `control_key`, silently labelling perturbed cells as controls (`control_key="DMSO"` flagged 3 of 4 keys including two real perturbations). It is a no-op when `group_cols` is empty. `align.py` needs no change — it reads from `filter` on a parallel branch and never sees composite keys.

## Backward compatibility

Defaults are inert and this is verified, not assumed. With `group_cols=[]`, `generate_feature_table.py` produces **byte-identical** outputs (sha256 on the aligned parquet, construct table and gene table, with and without pseudogene patterns), and `prepare_bootstrap_data.py` produces byte-identical `controls_arr`, `construct_features_arr`, `sample_sizes` and construct files. The `controls_arr` handed to `run_construct_bootstrap` is elementwise identical across HEAD, unwired, and explicit `scope="pooled"`.

## Tests

`tests/test_aggregate_grouping.py`, 10 tests, following the `test_phenix_wells.py` pattern. Each was verified to fail against `zarr3` by checking out the pre-change tree and shimming in the literal old implementations: 7 failed, and the remaining control-mask cases were shown non-vacuous by also failing a plausible wrong fix.

```
tests/test_aggregate_grouping.py            10 passed
tests/ (--ignore=tests/test_omezarr.py)     29 passed
```

The most subtle test guards index alignment: `join_well_annotations` must preserve the metadata index, because `merge` returns a fresh `RangeIndex` while the features frame keeps the original, and `features[mask]` aligns *on index*. Without it, cells silently pair with the wrong features.

`tests/test_omezarr.py` errors at collection on a pre-existing `ModuleNotFoundError: workflow.lib.shared.io`, unrelated to this change and left alone.

## Note for reviewers

Annotation columns must be declared metadata for every downstream script, since `split_cell_data` classifies anything absent from `metadata_cols` as a feature and then fails its dtype check. Seven scripts call `load_metadata_cols`, so rather than thread a parameter through each, the notebook appends the annotation columns to `METADATA_COLS` before writing `cell_data_metadata_cols.tsv` — the source of truth they all already read. `generate_feature_table.py` and `prepare_bootstrap_data.py` additionally add them defensively. If that TSV is hand-managed and stale, `align.py` fails loudly with `Invalid feature dtypes detected: ['treatment']`, which names its own fix.

Companion notebook parameters land in brieflow-analysis.

https://claude.ai/code/session_018JQtNBvewaEmdvYnC4k7Dm

### DIFF
--- a/tests/test_aggregate_grouping.py
+++ b/tests/test_aggregate_grouping.py
@@ -159,6 +159,22 @@ def test_control_mask_ignores_a_control_key_naming_a_group_value():
 # --- Bug class 3: well-annotation join -------------------------------------------
 
 
+def test_control_mask_list_survives_the_control_rename():
+    """prepare_alignment_data uniquifies controls to <name>_<pert_id>.
+
+    An exact-only list match stops seeing them after that rename, which silently
+    skips TVN normalization instead of failing. The suffixed form must still match,
+    while a longer name sharing the prefix (EGFP_10) must not.
+    """
+    values = pd.Series(
+        ["EGFP_1", "EGFP_1_CCTCCGGC", "H2B-EGFP_1_CCTCGCGT", "NES-EGFP_1", "EGFP_10"]
+    )
+
+    flagged = values[control_mask(values, ["EGFP_1", "H2B-EGFP_1"])].tolist()
+
+    assert flagged == ["EGFP_1", "EGFP_1_CCTCCGGC", "H2B-EGFP_1_CCTCGCGT"]
+
+
 def test_join_well_annotations_raises_on_unmapped_well(tmp_path):
     """An unmapped well becomes a NaN group that groupby drops without a word."""
     metadata = pd.DataFrame({"plate": [1, 1], "well": ["A1", "A2"], "cell": [10, 11]})
--- a/tests/test_aggregate_grouping.py
+++ b/tests/test_aggregate_grouping.py
@@ -77,6 +77,21 @@ def test_group_cols_empty_reproduces_single_key_aggregate():
     pd.testing.assert_frame_equal(meta, meta_default)
 
 
+def test_positional_call_still_binds_method_not_group_cols():
+    """group_cols must sit after method: 8_aggregate.py calls aggregate() positionally.
+
+    Inserting it before method silently bound AGG_METHOD ("median") to group_cols, and
+    list("median") splatted into single characters -> KeyError: 'm'.
+    """
+    metadata = pd.DataFrame({"gene": ["A", "A", "B", "B"]})
+    embeddings = np.array([[1.0], [3.0], [10.0], [30.0]])
+
+    _, meta = aggregate(embeddings, metadata, "gene", "median")
+
+    assert list(meta["gene"]) == ["A", "B"]
+    assert "group_cols" not in meta.columns
+
+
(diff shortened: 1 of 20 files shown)


mat10d:
Superseded by #258, which contains this branch in full and adds custom phenotype features, the reference_group control scopes, and the nanmedian fix. Closing so there is one PR to review.

PR #261, opened 2026-09-10
Add a per-cell num_nuclei column to the phenotype table

mat10d:
## What

Adds a `num_nuclei` column to the per-cell phenotype table: how many distinct segmented nuclei each final cell contains. The column is always present and defaults to 1, so existing configs pick it up with no changes.

## Why

Cells that contain more than one nucleus — genuinely binucleate cells, or two cells the segmentation merged — are currently indistinguishable from single-nucleus cells once the masks are reconciled. `reconcile_nuclei_cells` resolves the masks but records nothing about how many nuclei went into each final cell, so downstream analysis cannot filter them out or study them.

## How

Segmentation counts, for each final cell label, how many distinct pre-reconciliation nuclei fall inside it. Nuclei are first eroded to their center pixel (`center_pixels`, the same erosion `reconcile_nuclei_cells` already uses), so each nucleus is attributed to exactly one cell. `reconcile_nuclei_cells` itself is unchanged.

The counts are written per tile as a small `{cell, num_nuclei}` TSV next to the existing segmentation stats, joined onto the phenotype table in `extract_phenotype`, and carried through merge and aggregate.

`num_nuclei` is registered as reserved metadata, so aggregation treats it as a record rather than a feature: it is not fed into feature selection, filtering or perturbation scores. It behaves exactly like the existing reserved `offset_*` columns.

## Files

- `workflow/lib/shared/segmentation_utils.py` — new `count_nuclei_per_cell(nuclei, cells)` helper.
- `workflow/lib/shared/segment_cellpose.py`, `segment_stardist.py`, `segment_watershed.py` — keep a copy of the pre-reconcile nuclei mask, compute the per-cell counts, and return them from the `return_counts=True` cells branch.
- `workflow/scripts/shared/segment.py` — unpack the counts and write them as a fourth output; a header-only frame when cells are not segmented.
- `workflow/targets/phenotype.smk`, `workflow/targets/sbs.smk` — declare the new `nuclei_per_cell` TSV output. Both segment rules share `segment.py`, so both declare it; the SBS copy is written and not consumed.
- `workflow/rules/phenotype.smk` — feed the TSV into `extract_phenotype_cp`.
- `workflow/scripts/phenotype/extract_phenotype.py` — join the counts onto the phenotype table, defaulting to 1.
- `workflow/lib/aggregate/cell_data_utils.py` — add `num_nuclei` to `RESERVED_METADATA_COLS`.

## How it was verified

- `ruff format` and `ruff check` clean on all touched Python files.
- Synthetic label arrays: a cell containing two touching nuclei returns 2, a single-nucleus cell returns 1, under both `contained_in_cells` and `consensus` reconciliation.
- `snakemake -n` builds a DAG in both TIFF and OME-Zarr modes.
- `small_test_analysis` end to end, both modes (see Gate). `num_nuclei` is present with identical distributions in the phenotype, merge and aggregate outputs of both runs:
  - phenotype (per well, 1671 cells): 1637 x 1, 30 x 2, 2 x 3, 1 x 4, 1 x 8
  - merge (1612 cells): 1590 x 1, 20 x 2, 2 x 3
  - aggregate filtered (587 cells): 575 x 1, 10 x 2, 2 x 3
  - The small test reconciles with `contained_in_cells`, which keeps multi-nucleus cells, so the non-1 values are expected. A tile with no segmented objects writes a header-only TSV and gets an empty column.

## Gate

`tests/small_test_analysis` run locally on Slurm (16 cores, 64 GB), TIFF and OME-Zarr sequentially in one job: **both passed**, 3786/3786 steps, 57 min total.

An earlier run of the same gate caught a real bug — a tile with no segmented objects produces a phenotype frame with no columns at all, so indexing `label` raised a `KeyError`. That is fixed in the second commit and the rerun is green.

## Note for review

`format_singlecell_anndata.py` classifies `num_nuclei` into the h5ad `var` (features) rather than `obs`, because it does its own numeric/non-numeric split rather than consulting `is_reserved_metadata_col`. The existing reserved `offset_x`/`offset_y` columns land in `var` for the same reason, so this is pre-existing behaviour rather than something this change introduces, and I left it alone to keep the diff tight. Happy to route all reserved metadata columns to `obs` in a follow-up if that is the intent.

### DIFF
--- a/workflow/lib/aggregate/cell_data_utils.py
+++ b/workflow/lib/aggregate/cell_data_utils.py
@@ -137,8 +137,8 @@ def join_well_annotations(metadata, well_annotations_fp):
 
 
 RESERVED_METADATA_PREFIXES = ("offset_",)
-# per-cell list of secondary object labels; a record, not a feature
-RESERVED_METADATA_COLS = ("second_obj_ids",)
+# per-cell segmentation records (secondary object labels, nuclei count); not features
+RESERVED_METADATA_COLS = ("second_obj_ids", "num_nuclei")
 
 
 def is_reserved_metadata_col(col):
@@ -146,8 +146,8 @@ def is_reserved_metadata_col(col):
 
     Per-cell alignment-offset QC columns (offset_*) have screen-specific names (the
     alignment step/cycle count varies by screen), so they are matched by prefix rather
-    than enumerated in the metadata_cols file. Non-numeric per-cell records written by
-    secondary object detection are reserved by name.
+    than enumerated in the metadata_cols file. Per-cell records written by secondary
+    object detection and by segmentation are reserved by name.
     """
     return col in RESERVED_METADATA_COLS or col.startswith(RESERVED_METADATA_PREFIXES)
 
--- a/workflow/lib/shared/segment_cellpose.py
+++ b/workflow/lib/shared/segment_cellpose.py
@@ -40,6 +40,7 @@ from skimage.util import img_as_ubyte
 from skimage.segmentation import clear_border
 
 from lib.shared.segmentation_utils import (
+    count_nuclei_per_cell,
     image_log_scale,
     reconcile_nuclei_cells,
 )
@@ -200,7 +201,7 @@ def segment_cellpose(
     # Perform cell segmentation using Cellpose
     if cells:
         if return_counts:
-            nuclei, cells, seg_counts = segment_cellpose_rgb(
+            nuclei, cells, seg_counts, nuclei_per_cell = segment_cellpose_rgb(
                 rgb,
                 nuclei_diameter,
                 cell_diameter,
@@ -232,7 +233,7 @@ def segment_cellpose(
         print(f"Number of cells segmented: {counts['final_cells']}")
 
         if return_counts:
-            return nuclei, cells, counts_df
+            return nuclei, cells, counts_df, nuclei_per_cell
         else:
             return nuclei, cells
     else:
@@ -495,9 +496,12 @@ def segment_cellpose_rgb(
     )
 
     # Reconcile nuclei and cells if specified
+    raw_nuclei = nuclei.copy()
     if reconcile:
         print(f"reconciling masks with method how={reconcile}")
         nuclei, cells = reconcile_nuclei_cells(nuclei, cells, how=reconcile)
+
+    nuclei_per_cell = count_nuclei_per_cell(raw_nuclei, cells)
 
     counts["final_cells"] = len(np.unique(cells)) - 1
     print(
@@ -505,7 +509,7 @@ def segment_cellpose_rgb(
     )
 
     if return_counts:
-        return nuclei, cells, counts
+        return nuclei, cells, counts, nuclei_per_cell
     else:
         return nuclei, cells
 
--- a/workflow/lib/shared/segment_stardist.py
+++ b/workflow/lib/shared/segment_stardist.py
@@ -19,7 +19,7 @@ from stardist.models import StarDist2D
 from csbdeep.utils import normalize
 from skimage.segmentation import clear_border
 
-from lib.shared.segmentation_utils import reconcile_nuclei_cells
+from lib.shared.segmentation_utils import count_nuclei_per_cell, reconcile_nuclei_cells
 
 
 def segment_stardist(
@@ -95,7 +95,7 @@ def segment_stardist(
     # Perform cell segmentation using StarDist
     if cells:
         if return_counts:
-            nuclei, cells, seg_counts = segment_stardist_multichannel(
+            nuclei, cells, seg_counts, nuclei_per_cell = segment_stardist_multichannel(
                 dapi,
                 cyto,
                 model_type=model_type,
@@ -125,7 +125,7 @@ def segment_stardist(
         print(f"Number of cells segmented: {counts['final_cells']}")
 
         if return_counts:
-            return nuclei, cells, counts_df
+            return nuclei, cells, counts_df, nuclei_per_cell
         else:
             return nuclei, cells
     else:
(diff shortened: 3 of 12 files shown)


mat10d:
Can we also do this for SBS nuclei -> cells?

Would this collapse for either sbs or phenotyping if cell segmentation is turned off?

mat10d:
Both points addressed — details below.

### 1. SBS nuclei -> cells

`segment_sbs` already wrote the per-tile `nuclei_per_cell` TSV (it shares `scripts/shared/segment.py` with `segment_phenotype`); nothing consumed it. It is now joined into the SBS per-cell table the same way the phenotype side joins it into `phenotype_cp`:

- `workflow/rules/sbs.smk` — `rule extract_sbs_info` takes `SBS_OUTPUTS["segment_sbs"][3]` as a third input.
- `workflow/scripts/shared/extract_phenotype_minimal.py` — when that third input is present, map it onto the table's `cell` column, `.fillna(1).astype(int)`. The guard keeps `extract_phenotype_info`, which shares this script and passes only two inputs, unchanged.

`combine_sbs_info` uses `combine_dfs.py`, which concatenates whole frames, so the column carries into the per-well `sbs_info.parquet` without further changes. At aggregate, `num_nuclei` is already in `RESERVED_METADATA_COLS`, so it is treated as metadata rather than a feature.

small_test (both TIFF and OME-Zarr), values identical across modes:

```
sbs/tsvs/P-1_W-A1_T-0__sbs_info.tsv      3685 rows  num_nuclei: {1: 3524, 2: 149, 3: 11, 4: 1}
sbs/parquets/P-1_W-A1__sbs_info.parquet  6930 rows  num_nuclei: {1: 6716, 2: 196, 3: 17, 4: 1}
sbs/parquets/P-1_W-A2__sbs_info.parquet  6874 rows  num_nuclei: {1: 6636, 2: 209, 3: 25, 4: 4}
```

One thing worth flagging: the SBS `num_nuclei` stops at `sbs_info` and does not reach `merge_final`. The merge tables are fixed-column identity tables — `merge_triangle_hash` in `lib/merge/fast_merge.py` (and the equivalents in `lib/merge/merge.py` and `lib/merge/stitch_merge.py`) slices to `cols_final` = plate/well/tile/cell_0/i_0/j_0/site/cell_1/i_1/j_1/distance — and per-cell features join later in `final_merge` from the phenotype CP table only. So `merge_final.num_nuclei` is the phenotype value, and carrying the SBS one through would mean a second column (`num_nuclei_1`) added to three merge implementations. I left that out to keep this change small; happy to add it as a follow-up if you want it in the merged table. I did force a full merge + aggregate rerun in both modes to confirm the extra `sbs_info` column does not disturb alignment/merge: 12/12 (TIFF) and 28/28 (Zarr) steps green, `merge_final` unchanged at 1612 / 1611 rows per well.

### 2. Cell segmentation turned off

It does not collapse; `num_nuclei` is all 1. `scripts/shared/segment.py` writes the TSV unconditionally, header-only when `segment_cells` is false, so output[3] always exists and both joins map every label against an empty index and fall back to 1.

Dry runs, `snakemake -n --quiet rules` with `sbs.segment_cells: false` and `phenotype.segment_cells: false`, both DAGs build:

```
TIFF: 82 jobs   (segment_sbs 6, segment_phenotype 6, extract_sbs_info 6, extract_phenotype_cp 6, ...)
Zarr: 77 jobs   (same rules plus finalize_hcs_phenotype)
```

Real run with both flags off, TIFF, through `all_sbs all_phenotype` (merge needs cells, so it stops there) — completed, no errors:

```
nuclei_per_cell TSVs:  16 bytes each, header only ("cell\tnum_nuclei")
phenotype_cp (6 tiles):  3514 rows   num_nuclei: {1: 3514}
sbs_info    (6 tiles):  14648 rows   num_nuclei: {1: 14648}
phenotype_cp per-well parquets: 1753 / 1761 rows, all 1
sbs_info    per-well parquets: 7361 / 7287 rows, all 1
```

Normal gate on this branch, unchanged config: `run_brieflow.sh` and `run_brieflow.sh --zarr` both green.

mat10d:
Good catch — partially yes. `num_nuclei` was only covered by the `is_reserved_metadata_col` predicate; every explicit enumeration still omitted it. Now carried in:

- `workflow/lib/phenotype/constants.py` — `DEFAULT_METADATA_COLS`, next to the cell/nucleus/cytoplasm geometry entries. This is the canonical list the aggregate notebook writes out to `cell_data_metadata_cols.tsv`.
- `tests/small_test_analysis/config/cell_data_metadata_cols.tsv` — same position.
- `workflow/scripts/aggregate/format_singlecell_anndata.py` — it built `obs` from `metadata_cols` plus non-numeric columns and never consulted the predicate, so `num_nuclei` (and the pre-existing `offset_*`) were numeric and landed in `var` as if they were features. Its `obs` selection now also takes columns where `is_reserved_metadata_col` is true.

Already correct, no change needed: `lib/aggregate/cell_data_utils.py` `split_cell_data` (adds reserved columns to metadata via the predicate) and `RESERVED_METADATA_COLS` itself; `lib/aggregate/filter.py` and `lib/aggregate/perturbation_score.py` (both use the predicate to exclude from features).

I checked the other `DEFAULT_METADATA_COLS` consumers for the new entry:

- `lib/phenotype/extract_phenotype_cp_emulator.py` and `lib/phenotype/extract_phenotype_second_objs.py` use it only in `reorder_columns` — column ordering, no behavior change.
- `lib/shared/metrics.py` uses it twice as a subtract-metadata selector: `feature_count = len(all_columns) - len(metadata_cols_found)`, and dropping metadata before `f_classif` in the batch-effect metric. Both get slightly more correct, since `num_nuclei` was previously counted and tested as a feature. Neither iterates metadata columns expecting strings.

Gate green in both modes (`run_brieflow.sh`, `run_brieflow.sh --zarr`), and the single-cell h5ad now has these in `obs`:

```
ChCo-DAPI_COXIV_CENPA_WGA_CmCo-cell-nucleus-cytoplasm__singlecell.h5ad  (1362, 1633)
  obs: num_nuclei, offset_y, offset_x     var: (none)
  num_nuclei: {1: 1338, 2: 22, 3: 2}
ChCo-DAPI_CmCo-nucleus__singlecell.h5ad                                 (1086, 174)
  obs: num_nuclei, offset_y, offset_x     var: (none)
  num_nuclei: {1: 1066, 2: 17, 3: 3}
```

Identical for TIFF and OME-Zarr. `num_nuclei` is still metadata in the aggregate parquets (`features_singlecell.parquet`: 1069 / 1086 / 293 rows, values 1-3).

PR #262, opened 2026-09-10
perf(hcs): one pruned directory walk per plate store, parallel pixel passes

mat10d:
## Why

`finalize_hcs_preprocess_sbs` / `finalize_hcs_preprocess_phenotype` / `finalize_hcs_phenotype`
run for hours on real screens. Profiling the step showed three causes, all in
`workflow/lib/shared/hcs.py`:

1. **Eight to ten separate `Path.rglob("zarr.json")` walks per plate store.** `rglob` descends
   into every chunk directory of every multiscale level of every tile and every label
   (`{tile}/0/c/0/0/...`), so each walk is millions of stats on a network filesystem — repeated
   once per pass.
2. **The same zarr.json read and rewritten by several passes in sequence.** A label store's
   zarr.json was read three or four times and written up to three times.
3. **Every pixel read single-threaded.** The display histograms read the full level-0 array of
   every tile in the screen one at a time, and `segmentation_metadata` read every label array
   one at a time to count objects.

## What changed

**`workflow/lib/shared/hcs.py`** — new `_StoreIndex`: one `os.scandir` walk per store, pruned at
array directories (an array's zarr.json says `node_type == "array"`, so its chunk subtree is never
listed). It holds every parsed zarr.json keyed by relative path, and `flush` writes each changed
node back exactly once. `discover_plate_structure`, the `downsamplingMethod` re-injection, the
four label patches, histogram tile selection and window injection all run against that index.
The per-tile pixel passes (label object counts, display histograms) run over a thread pool sized
from `snakemake.threads`; histogram reads walk chunk-aligned row slabs so a whole tile is never
resident, and each worker merges its own batch. Above 256 tiles per store the histogram takes an
evenly spaced, deterministic subsample of that many tiles.

**`workflow/scripts/shared/write_hcs_metadata.py`** — passes `snakemake.threads` through to the
two library entry points.

**`workflow/rules/preprocess.smk`, `workflow/rules/phenotype.smk`** — `threads: 8` on the three
`finalize_hcs_*` rules.

iohub is kept for the per-position patches: it is ~10 s of the run on the benchmark fixture, so
removing it would buy little and risks changing what gets written.

Why a tile subsample and not a coarser pyramid level: the window is a display default and the
statistics are dataloader normalization constants, so an unbiased tile subsample is fine, but
gaussian downsampling would bias the standard deviation.

## Numbers

Benchmark fixture: one well (r02) of the zargun phenotype `aligned_1.zarr` — 4 wells, 277 tiles,
831 label stores, 2499 zarr.json, 1.2 GB. Slurm partition 20, 8 CPUs, 32 G.

| phase | before | after (8 threads) |
|---|---|---|
| `write_hcs_metadata` | 11.80 s | 2.51 s |
| `patch_store_metadata_with_iohub` | 64.90 s | 15.37 s |
| `compute_and_inject_omero_windows` | 31.30 s | 8.08 s |
| **wall clock (`/usr/bin/time -v`)** | **1:51.7** | **0:29.2** |
| peak RSS | 0.32 GB | 0.91 GB |
| `rglob` walks over the store | 9 | 0 |
| zarr.json reads | 4720 | 2782 |
| zarr.json writes | 1668 (up to 3 per node) | 1668 (one per node) |

Per sub-phase, before → after: `discover_plate_structure` 6.40 → 0 (folded into the index walk),
`_patch_label_versions` 4.36 → 0.00, `_patch_label_axis_units` 4.03 → 0.00, `_patch_label_scales`
3.76 → 0.00, `_patch_segmentation_metadata` 33.82 → 9.65, `_accumulate_channel_histograms`
27.04 → 9.39, `_inject_omero_windows` 4.53 → 0.53, iohub position loop ~10.1 → ~10.1.

zargun's `combined_metadata.parquet` has `pixel_size_x/y = NaN`, so `_set_per_dataset_scales` and
`_patch_label_scales` never fire on it. The same fixture with a pixel size filled in: 2:52.9 →
0:30.3 wall, zarr.json reads 5545 → 2782, writes 1939 → 1668.

The saving grows with screen size: the histogram pass is now bounded at 256 tiles per store
instead of every tile, and the walk cost no longer multiplies by the number of passes.

Peak RSS goes up rather than down, because the pixel passes are now concurrent. It is bounded by
thread count, not by screen size — 0.81 GB at `threads=4`, 0.91 GB at `threads=8`. Repeating the
old serial histogram loop over 2216 tile reads plateaus at 0.49 GB (0.411 / 0.446 / 0.479 / 0.491
GB and flat thereafter), so nothing in this step grows without bound with tile count; the 12 GB
seen in a `seff` report for `finalize_hcs_phenotype` is not reproducible at one-well scale and I
could not attribute it to code in `hcs.py`.

## Equivalence

Optimized run against a fresh fixture copy, every zarr.json diffed against the reference output of
the current code, normalized and compared as objects:

```
ref zarr.json: 2499   new zarr.json: 2499
STRUCTURAL DIFFS (everything except omero window/statistics): 0
WINDOW/STATISTICS DIFFS: 277 of 2499 files
   ch0 REF {"statistics": {"mean": 376.2607331453997, "median": 132.0, "std": 460.54114888721654}, "window": {"end": 2060.0, "max": 65535.0, "min": 0.0, "start": 45.0}}
   ch0 NEW {"statistics": {"mean": 375.6806820398234, "median": 132.0, "std": 460.8885192494333}, "window": {"end": 2062.0, "max": 65535.0, "min": 0.0, "start": 45.0}}
     window.end:      ref=2060.0000 new=2062.0000 delta=0.097%
     statistics.mean: ref=376.2607  new=375.6807  delta=0.154%
     statistics.std:  ref=460.5411  new=460.8885  delta=0.075%
   ch1 window.end:      ref=3832.0000 new=3819.0000 delta=0.339%
   ch1 statistics.mean: ref=395.1355  new=393.5401  delta=0.404%
   ch1 statistics.std:  ref=1043.1574 new=1041.6573 delta=0.144%
   ch2 window.end:      ref=8023.0000 new=8028.0000 delta=0.062%
   ch2 statistics.mean: ref=726.6598  new=726.1630  delta=0.068%
   ch2 statistics.std:  ref=1638.5519 new=1639.1374 delta=0.036%
RESULT: EQUIVALENT (window/statistics only)
```

`window.start` and `statistics.median` are unchanged on every channel; the largest difference
anywhere is 0.41 %. With the tile subsample disabled the two outputs are identical:

```
ref zarr.json: 2499   new zarr.json: 2499
STRUCTURAL DIFFS (everything except omero window/statistics): 0
WINDOW/STATISTICS DIFFS: 0 of 2499 files
RESULT: IDENTICAL
```

That second run also had pixel sizes filled in, so it covers the per-dataset and label coordinate
scales as well.

## Gate

`tests/small_test_analysis` in zarr mode (`bash run_brieflow.sh --zarr`, slurm partition 20,
16 CPUs, 64 G): **green**, `3786 of 3786 steps (100%) done`, elapsed 00:34:55, exit code 0, zero
`error in rule` lines. All three marker rules ran and their markers exist:

```
Finished jobid: 317 (Rule: finalize_hcs_phenotype)
Finished jobid: 337 (Rule: finalize_hcs_preprocess_sbs)
Finished jobid: 338 (Rule: finalize_hcs_preprocess_phenotype)

brieflow_output_zarr/phenotype/.hcs_done
brieflow_output_zarr/preprocess/.hcs_done_phenotype
brieflow_output_zarr/preprocess/.hcs_done_sbs
```

Its `finalize_hcs_phenotype` had `pixel_map entries=6`, so the gate also exercises the
per-dataset and label coordinate scale paths that zargun's NaN pixel sizes skip.

Tiff DAG untouched — `snakemake -n --quiet rules` in tiff mode, run from an isolated analysis
directory against both this branch and the base commit, produces byte-identical job tables (348
jobs, no `finalize_hcs_*` rules). The zarr-mode tables are byte-identical too, with the three
`finalize_hcs_*` rules present.

`ruff format` and `ruff check --line-length 100` clean on all four touched files;
`tests/test_phenix_wells.py` passes (21 tests), which covers `discover_plate_structure`.

### DIFF
--- a/workflow/lib/shared/hcs.py
+++ b/workflow/lib/shared/hcs.py
@@ -1,10 +1,15 @@
-"""HCS (High Content Screening) OME-NGFF metadata-only fusion utilities.
+"""OME-NGFF high-content-screening (HCS) metadata for brieflow plate stores.
 
-After Snakemake jobs write zarr stores directly into the HCS plate hierarchy
-(e.g., aligned_{plate}.zarr/{row}/{col}/{tile}/zarr.json), these functions
-discover what was written and layer the OME-NGFF metadata on top.
+Pipeline rules write zarr arrays straight into the plate hierarchy
+(``aligned_{plate}.zarr/{row}/{col}/{field}``). The functions here read back what
+was written and layer the OME-NGFF plate, well, field and label metadata on top:
+no arrays are moved, copied or symlinked, only ``zarr.json`` files are written.
 
-No symlinks or data copies — only zarr.json metadata files.
+The three public entry points run in order, once per plate store:
+``write_hcs_metadata`` (plate/row/well/labels groups), ``write_field_image_metadata``
+(per-field pixel scale, axis units, channel names, OMERO rendering, label
+annotations) and ``write_channel_intensity_statistics`` (screen-wide per-channel
+intensity percentiles and summary statistics).
 """
 
 import json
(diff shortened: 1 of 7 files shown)


workflow/scripts/shared/write_hcs_metadata.py
mat10d: This function seems inappropriately named, may be good to review this further

workflow/lib/shared/hcs.py
mat10d, line 1: hcs generally should be refactored to meet brieflow-style standards--regarding comments, how functions are organized in the script, and how functions docstrings look. The sectioned approach is not really up to par. Also, some function names may be inapprorpriate, can be more scientific. If there is any deadwood, let's clear now

workflow/rules/phenotype.smk
mat10d, line 1: Do we need to create an hcs for sbs data?

mat10d (commented):
Also--previous attempts to standardize to zarr3 format were approached here:

https://forum.image.sc/t/how-to-build-hcs-zarrs-with-multiple-image-types-per-fov/119329/10

https://github.com/ome/ngff/issues/476#issuecomment-5479051543

I am not sure if there are updates that in any way are material given what we chose. But just to keep in mind before a final pass.

mat10d:
Thanks for the review — all four points addressed. Two new commits: one for the module refactor and renames, one for the SBS HCS rule.

## 1. `workflow/lib/shared/hcs.py` refactored to brieflow style

The `# ---- section banner` comments are gone. The module now reads public API first — the three entry points in the order the script calls them — then the private helpers grouped by call order (store index → plate/well/field layout → per-field image metadata → label metadata → intensity statistics). Every docstring is Google style with a one-line summary and Args/Returns where there are arguments to describe, matching `file_utils.py` / `image_io.py`; all comments are single-line and at those files' density. The module is 1370 → 1215 lines.

**Public renames** (call sites updated in `write_hcs_metadata.py`):

| old | new |
|---|---|
| `patch_store_metadata_with_iohub` | `write_field_image_metadata` |
| `compute_and_inject_omero_windows` | `write_channel_intensity_statistics` |

`write_hcs_metadata` keeps its name — it does write the plate/row/well/labels group metadata, and it is what the script is named after.

**Private renames** — the "patch/inject/maybe" verbs are gone, and the names say what is being written rather than how:

`_discover_from_index` → `_find_fields`, `_maybe_write_labels_metadata` → `_write_labels_group`, `_is_label_store` → `_is_label_group`, `_zarr_v3_group_metadata` → `_group_metadata`, `_resolve_channel_names_for_store` → `_resolve_channel_names`, `_build_and_set_omero` → `_set_omero_rendering`, `_ensure_axes_units_micrometer` → `_set_axes_units_micrometer`, `_patch_label_versions` → `_set_label_versions`, `_patch_label_axis_units` → `_set_label_axis_units`, `_patch_label_scales` → `_set_label_scales`, `_build_segmentation_meta_for_label` → `_build_segmentation_metadata`, `_patch_segmentation_metadata` → `_write_segmentation_metadata`, `_tile_channel_histograms` → `_field_channel_histograms`, `_stats_from_histograms` → `_statistics_from_histograms`, `_inject_omero_windows` → `_write_intensity_metadata`, `_MAX_HISTOGRAM_TILES` → `_MAX_HISTOGRAM_FIELDS`.

`_StoreIndex` stays, but its methods now read as plate/well/field/label operations: `tiles()` → `fields()`, `mark()` → `mark_dirty()`, `refresh()` → `reload()`. "field" rather than "tile" throughout, since that is the OME-NGFF term for what sits under a well.

**Deadwood removed:**

- `discover_plate_structure` — a one-line public wrapper around the private discovery pass with no pipeline callers left after the index landed. Its only remaining user was the Phenix regression test from #249, which now calls `_find_fields(_build_store_index(plate))` directly, so the coverage is unchanged (`tests/test_phenix_wells.py`, 21 passed).
- `_split_well` — a deprecated shim that just forwarded to `file_utils.split_well`, plus the now-unused import.
- The `rebuild=` flag on `_get_store_index`, replaced by an explicit `_build_store_index` for the one caller that wants a fresh walk.
- Debug prints: `[patch] store=`, `[patch] pixel_map entries=`, `[patch] positions=`, `[patch] skipping unexpected pos_path=`, `[patch] segmentation_metadata set on N`, `[iohub patch] metadata root missing`. What is left is one progress line per store plus the per-channel window/statistics summary, which is the density the other scripts print at.

No behavior change — see the equivalence numbers at the bottom.

## 2. `write_hcs_metadata.py:16`

`compute_and_inject_omero_windows` → `write_channel_intensity_statistics`. It accumulates a full-precision uint16 intensity histogram per channel across every image field of every store in the screen, then writes the 1st/99th percentiles as the OMERO display window and the mean/std/median as `statistics` on each field's `omero.channels[i]`. The name now says that instead of naming the OMERO key and the verb.

## 3. `rules/phenotype.smk` — "Do we need to create an hcs for sbs data?"

Yes, and there is now a `finalize_hcs_sbs` rule mirroring `finalize_hcs_phenotype`: marker `sbs/.hcs_done`, `channels_metadata` from `preprocess.sbs_channels_metadata`, `channel_names` from `sbs.channel_names`, `modality="sbs"`, `threads: 8`, wired into `all_sbs` the same way.

What it covers: in zarr mode `targets/sbs.smk` sets `_sbs_img_temp = None`, so **every** SBS image store survives, not just `aligned`. The rule therefore takes all six — `aligned`, `illumination_corrected`, `log_filtered`, `standard_deviation`, `peaks`, `max_filtered` — plus the `labels/nuclei` and `labels/cells` stores nested under each field.

The parsing helpers needed no changes: `_parse_store_type` already handles the multi-word names (`standard_deviation_1.zarr` → `standard_deviation`), `_infer_modality_from_store_path` picks up `sbs` from the path, `peaks` and `standard_deviation` were already in `_SINGLE_CHANNEL_STORES`, and the pixel-size lookup reads `preprocess/metadata/sbs/{plate}/{row}/{col}/combined_metadata.parquet`, which preprocess writes for SBS exactly as it does for phenotype. `sbs` config carries the `nuclei_diameter` / `cell_diameter` / `*_flow_threshold` / `*_cellprob_threshold` / `dapi_index` / `cyto_index` keys the segmentation metadata builder wants, so the label stores get the same provenance block phenotype gets.

**Channel naming did need a change.** `align_cycles` returns `(cycle, channel, y, x)` and `save_image` maps a 4-D array to `TCZYX` as `(C, Z, Y, X)` — so in the SBS `aligned`, `log_filtered` and `max_filtered` stores the OME **channel axis holds sequencing cycles and the stains sit on Z**. Confirmed on the small test data: `aligned_1.zarr/A/1/0/0` has shape `[1, 11, 5, 1200, 1200]` — 11 cycles on C, 5 stains on Z — while `illumination_corrected_1.zarr` (written before cycle stacking) is `[1, 5, 1, 1200, 1200]`, a genuine 5-channel image.

Passing `sbs.channel_names` blindly would therefore be a live hazard: the existing rule is "use the config names if the count matches", and a screen with 5 cycles and 5 channel names (`DAPI, G, T, A, C` — the default) would have its cycles silently labelled as bases. So `_resolve_channel_names` now takes the modality, and for SBS cycle-stacked stores names the channels `cycle_1 … cycle_N`. `illumination_corrected` still gets the real stain names from `sbs.channel_names` by the normal count-match rule, and `peaks` / `standard_deviation` are still named after the store. Phenotype behavior is untouched.

Consequence worth knowing: for the cycle-stacked SBS stores the per-channel display window and normalization statistics are computed **per cycle**, which is what you want for rendering, and the stains are the Z slider.

## 4. Upstream: image.sc thread and ome/ngff#476

Read both. Short answer: **nothing there changes the layout this module writes, and nothing needs to change in this PR.**

- The image.sc thread ends at your own post 10 choosing Option D — one spec-compliant HCS plate store per image type, labels under the primary image, a lightweight manifest for co-registration. The replies backing it are unchanged, and jluethi's Fractal alternative (Option F, derived images as extra `images[]` entries in the well) is explicitly the one that does not scale to our 100s–1000s of fields per well. d-v-b's point about not nesting `.zarr`-suffixed directories is already honoured — our label groups are `labels/nuclei` and `labels/cells`, no suffix.
- ome/ngff#476 ("Allow nesting HCS and Scenes?") is still **open** and has not moved toward anything we could adopt. The August 31 comments are the newest: LucaMarconato and jo-mueller converge on Option 2 (`plate` + `scene` metadata coexisting at the top) rather than the nesting they originally preferred, and will-moore's comment 5479051543 closes it out with "I don't see any changes happening now before v1.0 (RFC-8)", pointing at the RFC-8 HCS section, whose well layouts nest by `acquisition` and mention no transforms. jluethi's position is that this belongs with Collections, and that until then the important thing is that every individual image is a spec-compliant OME-Zarr — which Option D gives us by construction.
- The one genuinely new thing since we chose is that RFC-5 now names the "Scene" layout, and 0.6 has landed, so Option G (a scene inside the field group with identity transforms linking the image types) is no longer theoretical. That is the eventual answer to the co-registration gap in Option D, but it needs the well spec to allow a scene where it currently requires a multiscales group — which is exactly what #476 is asking for and what will-moore says is not happening before v1.0.

**Follow-on, not this PR:** once wells may contain scenes, the migration from Option D is additive — write a `scene` block with identity transforms between the per-type stores, without restructuring any of them. Worth revisiting when RFC-8/Collections lands; I'd rather not build a custom scene block against a spec that is still moving.

## Verification

All on `zargun` phenotype `aligned_1.zarr` well `r02` (277 fields, 831 label stores, 2499 `zarr.json`, 1.2 GB), Slurm partition 20, 8 cpus, 32 G — the same fixture the original numbers were taken on.

**Equivalence.** Refactored vs the pre-refactor head of this branch, same input, same threads:

```
A=run_prehead/phenotype/aligned_1.zarr  (2499 zarr.json)
B=run_opt/phenotype/aligned_1.zarr  (2499 zarr.json)
only in A: 0  only in B: 0
identical: 2499
structural diffs: 0
differ only in omero window/statistics: 0
max relative delta in window/statistics: 0.0000 %  None
```

Byte-identical — the refactor is behavior-preserving. Against the pre-PR baseline (`84e5544`, serial, no subsampling), unchanged from the original claim:

```
A=run_ref/phenotype/aligned_1.zarr  (2499 zarr.json)
B=run_opt/phenotype/aligned_1.zarr  (2499 zarr.json)
only in A: 0  only in B: 0
identical: 2222
structural diffs: 0
differ only in omero window/statistics: 277
max relative delta in window/statistics: 0.4038 %  ('r02/c02/1/zarr.json', 'statistics.mean', 395.1354889121639, 393.54011527858154)
```

**Timing** — no regression from the refactor:

| phase | baseline `84e5544` | this branch, pre-refactor | this branch, refactored |
|---|---|---|---|
| `write_hcs_metadata` | 6.53 s | 2.20 s | 2.27 s |
| field image metadata | 53.16 s | 13.59 s | 11.88 s |
| channel intensity statistics | 25.75 s | 6.91 s | 7.37 s |
| **total** | **85.44 s** | **22.70 s** | **21.53 s** |
| `/usr/bin/time` wall | 1:28.74 | 0:26.02 | 0:25.00 |
| peak RSS | 0.32 GB | 0.89 GB | 0.89 GB |
| `zarr.json` reads / writes | 4720 / 1668 | 2782 / 1668 | 2782 / 1668 |
| `rglob` calls | 9 | 1 | 1 |

**Dry runs.** `snakemake -n --quiet rules` in tiff mode is byte-identical to the pre-refactor checkout — 348 jobs, zero `hcs` rules. Zarr mode now lists `finalize_hcs_sbs` alongside the other three.

**Small test gate**, `bash run_brieflow.sh --zarr` from a wiped `brieflow_output_zarr/`, one sbatch on partition 20 (16 cpus, 64 G):

```
3787 of 3787 steps (100%) done
Complete log(s): .../.snakemake/log/2026-09-10T090514.488286.snakemake.log
```

`COMPLETED`, 33:29 elapsed, zero `Error in rule` / `MissingOutputException`. All four HCS rules ran:

```
Finished jobid: 338 (Rule: finalize_hcs_preprocess_sbs)
Finished jobid: 339 (Rule: finalize_hcs_preprocess_phenotype)
Finished jobid: 345 (Rule: finalize_hcs_sbs)
Finished jobid: 349 (Rule: finalize_hcs_phenotype)
```

`finalize_hcs_sbs` output, showing the six stores and the per-cycle channels:

```
Writing HCS metadata for: brieflow_output_zarr/sbs/aligned_1.zarr
  aligned_1.zarr: wrote 18 zarr.json
Writing HCS metadata for: brieflow_output_zarr/sbs/illumination_corrected_1.zarr
  illumination_corrected_1.zarr: wrote 6 zarr.json
Writing HCS metadata for: brieflow_output_zarr/sbs/log_filtered_1.zarr
  log_filtered_1.zarr: wrote 6 zarr.json
Writing HCS metadata for: brieflow_output_zarr/sbs/standard_deviation_1.zarr
  standard_deviation_1.zarr: wrote 6 zarr.json
Writing HCS metadata for: brieflow_output_zarr/sbs/peaks_1.zarr
  peaks_1.zarr: wrote 6 zarr.json
Writing HCS metadata for: brieflow_output_zarr/sbs/max_filtered_1.zarr
  max_filtered_1.zarr: wrote 6 zarr.json
HCS metadata written for 6 plate zarr(s).
Computing screen-wide channel intensity statistics [1.0, 99.0]th pct over 6 store(s)...
  channel 0: window=(0.0, 4868.0)  mean=496.3 std=1013.4 median=78.0
  ...
  channel 10: window=(0.0, 4689.0)  mean=522.3 std=1050.3 median=97.0
Channel intensity statistics written to 36 zarr.json files.
```

Eleven channels because the small test has eleven cycles, as expected. `pytest tests/` (excluding small_test_analysis): 124 passed, 5 skipped.

## One follow-on worth deciding, not changed here

The intensity statistics pass pools histograms across every store the rule is given, then applies one window per channel index to all of them. For phenotype that pools `aligned` and `illumination_corrected`, which are the same pixels at two stages — fine. For SBS it now pools six stores of quite different semantics, so channel 0's window mixes `aligned` cycle 1 with `peaks` and `standard_deviation`, which are single-channel derived images on a different scale. The window is only a display default and the statistics are normalization constants, so nothing is broken, but "one window per store type" is probably the more defensible grouping. I left the pooling alone rather than change phenotype's behavior in a PR whose claim is that behavior is unchanged — happy to do it as a follow-up if you agree it should be per store type.

workflow/lib/shared/hcs.py
```diff
@@ -1,398 +1,420 @@
...
+    },
+}
+
+_AXIS_UNITS = {"X": "micrometer", "Y": "micrometer", "Z": "micrometer", "T": "second"}
 
-# ---------------------------------------------------------------------------
-# High-level API (used by Snakemake scripts)
-# ---------------------------------------------------------------------------
+# Cap on fields histogrammed per store; above this an evenly spaced subsample
+# is used for the intensity percentiles and summary statistics.
```
mat10d: Multiline comments are still banned--this should be reflected. inthe skill as well

PR #264, opened 2026-09-10
Imaris file reader (lands #193)

mat10d:
# Description

Lands @joandrews33's Imaris (`.ims`) reader from #193 on top of current `main`, plus a small follow-up
tidy-up. All of Owen's commits are contained here with his authorship intact (this branch is
`main` merged with `joandrews33:ims-reader`), so merging this PR will also mark #193 as merged.

#193 itself is untouched — it was opened in February and had fallen behind `main`, which is why the
`Test Analysis` gate never ran on it. This branch exists so the gate can run.

## What came from #193

- `imaris-ims-file-reader==0.1.8` added to `pyproject.toml`.
- `extract_metadata_ims` / `convert_ims_to_array` in `workflow/lib/preprocess/preprocess.py`, plus
  `"ims"` in `DATA_FORMATS` and the `extract_metadata` / `convert_to_array` dispatch.
- A one-line comment in `workflow/scripts/preprocess/extract_metadata.py`. The legacy
  metadata-file behaviour there is unchanged.

The merge was clean — `preprocess.py` had not moved on `main` since the PR branch point.

## Follow-up commits in this PR

- **Tidy Imaris reader helpers**
  - Lifted the three nested HDF5 helpers out of `_extract_ims_per_file_metadata` into module-level
    private helpers, with the public `extract_metadata_ims` first (house style).
  - Optional `DataSetInfo` attributes now go through a coercion helper: a missing
    `XPosition`/`BinningX`/`LensPower` yields `None` instead of raising on `float(None)`.
  - `DataSetInfo/CustomData` and `DataSetInfo/Image` are looked up defensively, so a file without
    those groups no longer raises `KeyError`.
  - Tile parsing uses a regex on the trailing `F<number>` of the file stem (this puts the
    previously unused `re` import to work). It is `Path`-safe, and falls back to the tile passed by
    the caller when the file name has no such field, rather than raising.
  - The Imaris `FieldID` now lands in its own `field_id` column, and `well` comes from the caller
    when given (falling back to `FieldID`). `combine_metadata` filters on `well` against the
    configured well, so a `FieldID` there would have dropped every row.
  - `cycle` and `round` are passed through to the output frame the way the ND2 and TIFF extractors
    do, and the mm-to-µm rescale is applied once on the column instead of per row.
  - `convert_ims_to_array` gained a `**kwargs` tail and `convert_to_array` forwards `**kwargs` to
    it, matching the ND2/TIFF converters.
  - Documented why `flip_y_axis` defaults to `True`.
- **Document `.ims` among the supported preprocessing inputs** (`docs/source/index.md`).

## Dependency check

`imaris-ims-file-reader==0.1.8` is a 15 kB pure-Python wheel requiring `numpy`, `h5py>=3.5.0`,
`scikit-image` and `zarr`, all unpinned. Resolving the full `pyproject.toml` dependency set for
linux/py3.11 with and without it gives the same result: `numpy` stays at the pinned `2.0.2`,
`h5py` (3.16.0) was already pulled in by `tensorflow`/`keras`/`microfilm`, and `zarr` by
`dexp`/`ome-zarr`/`spotiflow`. No new pin is needed and `h5py` does not have to be declared
explicitly.

## Verification

- `ruff format` + `ruff check --line-length 100` clean.
- Synthetic `.ims` round-trip (built with `h5py` in the Imaris on-disk layout): 3 channels ×
  2 z-planes; `extract_metadata` returns the expected positions/binning/magnification and
  `convert_to_array` reproduces the max projection with the Y flip. Deleting `LensPower` and
  `XPosition` from the file now returns `None`/`NaN` instead of raising.
- `snakemake -n` over all six `all_*` targets on the small test data: 333 jobs, identical to
  `main` — no rules or targets are touched by this change.

## What is left for the author

Owen noted in #193 that he had prototyped letting `extract_metadata.py` use a dedicated metadata
file *and* per-image bundled metadata at the same time (Imaris has both per-image and per-job
metadata), but could not wire it into Snakemake without deeper structural changes, so he reverted
to legacy behaviour. That is still open and worth doing as its own PR — it would need a rule-level
change so `extract_metadata` can take both `samples` and `metadata` inputs at once, rather than
choosing between them.

Also still unverified without real data: whether `flip_y_axis=True` and the mm→µm
`rescale_positions=1000` default are right for every Imaris acquisition, and whether the
`F<number>` tile convention holds beyond Owen's file naming. All three are now parameters with
documented defaults rather than hard assumptions.

Enhancement (adds functionality)

# Checklist

- [x] I have updated the `pyproject.toml` to reflect the change as designated by [semantic versioning](https://semver.org/) (1.4.14 -> 1.5.0).
- [x] I have made corresponding changes to the documentation.


mat10d:
Superseded by #265, which lands the same commits (Owen's #193 plus the tidy fixes) on zarr3 instead of main, since zarr3 merges into main next week. Branch deleted.

PR #268, opened 2026-09-11
perf: snakemake-path performance optimizations

petercla0119:
# perf: snakemake-path performance optimizations

Performance optimizations for the Snakemake execution path. Change preserves upstream behavior on both the Snakemake and direct-runner paths; where an intermediate format changes (TSV→parquet) upstream semantics (`num_nuclei` enumeration, secondary-object handling) are preserved.

## Why

On real screens (plate-scale, ~1M cells/well, ~3600 phenotype columns) several steps were the wall-clock bottleneck or OOM'd outright. These changes make the pipeline finish where it previously stalled, without altering scientific output (two intentional exceptions called out below).

Note that some changes led to a larger disk storage space footprint - refer to `*.smk` files due to intermediate/temp files being retained.

## What changed — before/after

| Area | Change | Before → After |
|---|---|---|
| **merge** | `final_merge` streams the wide phenotype join out-of-core with polars `scan_parquet → left join → sink_parquet` instead of a pandas full-load merge that OOM'd | **721s → 44s** (well A1, ~3600 cols × 1.06M rows) |
| **merge** | `fast_merge` groups tiles once instead of per-pair | **O(n²) → O(n)** |
| **segmentation** | `identify_cytoplasm` per-region loop → vectorized `np.where` (bit-identical masks) | **~357×** |
| **segmentation** | `lru_cache` the Cellpose model (built once per worker, not per tile), eval `batch_size=32`, multi-GPU `select_gpu_device` fan-out | model build amortized; GPU fan-out |
| **io** | Per-tile phenotype tables, sbs reads/`sbs_info`, and combine step migrated TSV→parquet via shared `parquet_io` + `combine_dfs` | write traffic **5.0× lower**, on-disk **2.7× smaller**, merge read-back **1.9× faster** |
| **sbs** | `align_sbs` error-correction uses an O(1) hamming-1 lookup index instead of O(n·library) scan; `call_cells`/`eval_mapping` project only needed columns and skip unused QC recompute | O(1) correction lookup |
| **phenotype** | `extract_phenotype_cp_emulator` gains per-object parallelism (`n_jobs`); callers cap BLAS threads so nested threadpools don't oversubscribe cores | parallel extraction, no oversubscription |
| **preprocess** | `calculate_ic_field` gains a tunable `n_jobs` (config `preprocess.ic_n_jobs`, default 8): batched memory-bounded accumulate reads + `applyIJ_parallel` median filter. **Bit-identical** to the serial path (read order + float summation order preserved) | parallel IC, OOM-free |
| **aggregate** | `generate_feature_table` computes construct-level medians lazily via `scan_parquet().group_by().median()` instead of accumulating every cell's feature vector across batches | O(1) memory aggregation |
| **scheduling** | `threads`/`resources`/`gpu` directives on sbs/phenotype/merge rules; pre-segmentation intermediates no longer marked `temp()` so a rerun before segmentation doesn't recompute the whole pre-seg chain (18 sbs + 15 phenotype intermediates); preprocess reads ND2 binning from the text-info block only (no multi-GB pixel materialization) | correct parallelism, no pre-seg recompute sawtooth |

## Behavior changes (intentional — please review)

1. **Aggregate pseudo-gene grouping (see #258).** The lazy polars aggregation folds group labels into the perturbation id, so grouped construct medians are preserved. The one change: pseudo-gene grouping now runs over the flat construct table rather than being split per group label. This is the most likely point of pushback — flagging it explicitly.
2. **IC-field `n_jobs` default.** The `calculate_ic_{sbs,phenotype}` rules now carry `threads: config.get("preprocess",{}).get("ic_n_jobs", 8)`; upstream previously hardcoded `n_jobs=-1` inside the function. Output is bit-identical; only the scheduler thread reservation changes.

## Deliberately excluded (separate future PRs)

- **IC-field reproducibility seeding** — `calculate_ic_field` subsamples tiles with an unseeded RNG. A `random_seed` knob will land as its own PR (kept separate from the parallelization here).
- **RANSAC determinism seeding** — seeding `RANSACRegressor` in the merge alignment path will land as its own PR.

## Testing

Full E2E on the small test dataset passed clean end-to-end (**3797/3797 steps, exit 0**), exercising the parquet migration and the polars aggregate path; parquet outputs verified (phenotype_cp, aggregate feature tables, `merge_final.parquet`) with no stray TSVs. Unit tests: 124 passed, 5 skipped.

Performance optimizations in this PR have not been tested with SLURM.

### DIFF
--- a/workflow/Snakefile
+++ b/workflow/Snakefile
@@ -26,6 +26,7 @@ wildcard_constraints:
     # Accept both plate-well style (row=letters "A", col=digits "1") AND Opera Phenix
     # numeric wells (row="r02", col="c02"). Segments are "/"-delimited in paths so
     # widening these does not introduce cross-wildcard ambiguity.
+    well=r"[A-Z][0-9]+",
     row=r"[A-Za-z]+\d*",
     col=r"[A-Za-z]*\d+",
     tile=r"[0-9]+",
--- a/workflow/lib/external/cp_emulator.py
+++ b/workflow/lib/external/cp_emulator.py
@@ -627,17 +627,27 @@ def lstsq_slope_all_multichannel(r):
 def cp_colocalization_all_channels(r, mode="multichannel", **kwargs):
     if mode == "multichannel":
         channels = r.intensity_image.shape[-1]
-    else:
-        channels = r.intensity_image_full.shape[-3]
-
-    results = [
-        cp_colocalization(r, first, second, mode=mode, **kwargs)
-        for first, second in combinations(list(range(channels)), 2)
-    ]
-
-    if mode == "multichannel":
+        # Precompute otsu threshold and rankdata once per channel instead of per pair.
+        # Each channel appears in (C-1) pairs; caching cuts calls from 2*C(C,2) → C
+        # (e.g. 12 → 4 for C=4). Algebraically identical to the per-pair computation.
+        V = r.intensity_image[r.image]  # (n_pix, C) masked pixel array
+        _thr = [otsu(V[:, c]) for c in range(channels)]
+        _ranks = [rankdata(V[:, c], method="dense") for c in range(channels)]
+        results = [
+            measure_colocalization(
+                V[:, first], V[:, second],
+                _A_thresh=_thr[first], _B_thresh=_thr[second],
+                _A_ranks=_ranks[first], _B_ranks=_ranks[second],
+            )
+            for first, second in combinations(range(channels), 2)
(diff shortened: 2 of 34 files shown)


petercla0119:
Superseded by #283 — same head commit, no content change. Reopened only to move the head branch under an `up/` prefix so upstream-bound branches are distinguishable from this fork's local development branches. Closing here; please review #283.

PR #263, opened 2026-09-10
Make the Streamlit visualizer read OME-Zarr pipeline output

mat10d:
The `visualization/` Streamlit app was written against the TIFF/TSV output layout, so pointing it at a `--zarr` run showed empty filters, untitled plot groups and no montages. This makes it format-aware so one app serves both output modes.

## What zarr mode actually changes

Only the *path shape* changes for preprocess/sbs/phenotype outputs — eval plots and tables stay PNG/TSV, and merge/aggregate/cluster outputs keep flat `get_filename` names in both modes. The one real format change is montages.

## Audit

| Assumption in the visualizer | zarr3 reality | Fix |
|---|---|---|
| `CONFIG_PATH` is a single YAML file | the runner passes `config/config.yml config/config_zarr.yml` to `snakemake --configfile`, which deep-merges them; the visualizer only read the first, so it never saw `image_format: zarr` | `src/config.py` splits `CONFIG_PATH` and deep-merges in order, mirroring snakemake; adds `get_image_format()` |
| `FileSystem.extract_plate_id` reads a `P-\d+` filename prefix | `get_data_output_path(..., img_fmt="zarr")` nests the location as directories: `sbs/eval/mapping/1/mapping_overview.tsv` | falls back to `parse_nested_location()`, which recovers the location chain from the directories below an `eval`/`tsvs`/`parquets` anchor |
| `FileSystem.extract_well_id` reads a `W-[A-Z]\d+` prefix | zarr splits the well into `row`/`col` directories (`.../1/A/1/0/bases.tsv`) | same fallback, rejoining `row + col` |
| `FileSystem.extract_metric_name` returns the part after the last `__` | zarr basenames are the bare metric (`mapping_overview.tsv`), so every metric came back `None` and the Metric filter was empty | falls back to the filename stem |
| `VisualizationRenderer` titles groups from `parse_filename` attrs | attrs are empty in zarr mode, leaving a dangling `"Metric - "` title | reuses `parse_nested_location`, and drops the separator when there are no attrs |
| `FileSystem.find_files` matches files by extension | a `.zarr` store is a *directory*, and a `.file_manifest` only lists the chunk files inside it | `extensions=["zarr"]` now resolves stores, collapsing manifest entries back to their innermost `.zarr` ancestor so the manifest fast path still works |
[...]

PR #91, opened 2025-04-15
Spotiflow

mat10d:
# Description

Thank you for your contribution to Brieflow!
Please _succinctly_ summarize your proposed change.
What motivated you to make this change?

Please also link to any relevant issues that your code is associated with.

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [X] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [X] My code follows the style guidelines of this project.
- [X] I have checked linting and formatting with `ruff check` and `ruff format`.
- [X] I have performed a self-review of my own code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have made corresponding changes to the documentation.
- [X] My changes generate no new warnings.
- [X] I have deleted all non-relevant text in this pull request template.

PR #271, opened 2026-09-14
Run mozzarellm cluster annotation as a pipeline module

mat10d:
Builds the mozzarellm annotation step into the pipeline as a two-rule module, and fixes the screen context it feeds the model.

Branched off `zarr3-e2e` (#263), so this PR contains that branch's commit as well as its own; it is meant to land after #263 merges into `zarr3`.

## The module

`rule annotate_clusters` — one job per row of the mozzarellm combo table. It reads that clustering's `cluster.h5ad` and the row's screen-context JSON, and writes

```
cluster/<channel_combo>/[<compartment_combo>/]<cell_class>/<leiden_resolution>/
  mozzarellm/<run_name>/<screen_name>_clusters.json
  mozzarellm/<run_name>/<screen_name>_genes.csv
  mozzarellm/<run_name>/traces/…
```

The run directory is named, not timestamped, so snakemake can name the outputs. The job always calls `run_mozzarellm(..., resume=True)`, so a retry after a transient provider failure re-reads from the traces the clusters that already answered rather than paying for them again — which is why the rule also carries `retries: 2`. `threads:` comes from `max_workers` and memory is 4G; these jobs are API-bound.

`rule annotate_cluster_anndata` — reads the clustering h5ad plus the two files above and writes `<resolution_dir>/cluster_annotated.h5ad`:

- `obs["mozzarellm_category"]`, `obs["mozzarellm_subclass"]`, `obs["mozzarellm_rationale"]` — the per-gene call, empty string for perturbations the run left unclassified.
- `uns["mozzarellm"]` — `run_name`, `screen_name`, `leiden_resolution`, `cluster_column`, and `clusters`, a mapping from cluster label to `{dominant_process, pathway_confidence, summary}`. `cluster_column` names the `cluster_group_<res>` obs column whose values key `clusters`, so a reader goes perturbation → cluster label → call without touching the run directory.

The merge itself lives in `workflow/lib/cluster/mozzarellm_io.annotate_cluster_anndata`; both scripts are thin.

## Config contract

```yaml
mozzarellm:
  combo_fp: "config/mozzarellm_combo.tsv"
  run_name: "run1"
  screen_name: "<screen>"
  model: "claude-sonnet-5"
  mode: "cot"
  mcp: true
  include_features: "auto"
  include_strength: "auto"
  n_features: 5
  fdr_threshold: null
  max_tokens: 64000
  max_workers: 8
```

`mozzarellm_combo.tsv`: `cell_class`, `channel_combo`, `leiden_resolution`, `screen_context_fp`, plus `compartment_combo` when the screen splits by compartment. It is the selection of which clusterings get annotated — deliberately a subset of `cluster_combo.tsv`, not a copy of it.

The module is gated on `mozzarellm.combo_fp` being set rather than on the block existing, because the `mozzarellm:` block is already claimed by the visualizer for its default-view keys (`cell_class` / `channel_combo` / `leiden_resolution`, documented in `docs/source/3.running_modules.md`). Keying off block presence would have crashed every screen that already carries those three keys. Both sets of keys coexist in one block.

## The screen-context fix

`screen_context_from_screen` built `target_phenotype` from `config["phenotype"]["channel_names"]` — the screen's full panel — but a clustering is per channel combo, so a CENPA-only clustering was telling the model the features were measured across all four channels. The context now names only the combo's channels, each with its `marker_of` text from `screen.yaml`'s `phenotype.channels`, and names the cell class when it is not `all`:

```
before: … measured across the DAPI, COXIV, CENPA, WGA imaging channels
after:  … features of Interphase cells measured in the DAPI (marker of Nuclei) imaging channel
```

`screen_context_for_combo(screen, config, channel_combo, cell_class, leiden_resolution)` is the public entry point for the marimo notebook, which pre-writes one context JSON per table row. `screen_context_from_screen` keeps working unchanged for callers that pass no combo. The context also carries `organism_ontology_term_id`, so a job reading a written context can resolve the same UniProt taxonomy id the screen description would have given without reading `screen.yaml`; `mozzarellm`'s context schema allows extra keys.

`run_mozzarellm` gains `run_name`, `screen_context`, `organism_id` and `max_workers`. `max_workers` is passed only when the installed `analyze_screen` accepts it (it does not yet on the pinned commit), so this does not break against the current package.

## Visualizer

The cluster page now prefers `cluster_annotated.h5ad` and falls back to `mozzarellm/run_*/<screen>_clusters.json`, so screens annotated before this stage still render. The run-directory glob is no longer `run_*`-only, and the helpers key off the resolution directory rather than the run directory.

## Install recipe

`docs/source/2.installation_analysis_setup.md` said `uv pip install -e .`, which takes core deps only — that is why no screen environment has mozzarellm. It now installs `-e ".[mozzarellm]"`. It stays an extra; nothing moved into core dependencies.

## What was run

`zarr3` has no CI, so the evidence is dry-runs and targeted exercises of the new code paths, from a scratch analysis dir over the small-test config (`split_by_compartment: true`, so the compartment segment is exercised):

- **Inert with no config block.** `snakemake -n --quiet rules --until all_preprocess … all_cluster`, this branch vs. its base, tiff and zarr: job tables are byte-identical in both modes.
- **With a block and a two-row table** (a subset of the three-row `cluster_combo.tsv`), tiff and zarr: 2 × `annotate_clusters` + 2 × `annotate_cluster_anndata` + `all_mozzarellm`, on the right paths (`cluster/DAPI/nucleus/Interphase/2/mozzarellm/run1/…`), each job taking its own row's context JSON, `threads: 8`, `mem_mb=4000`.
- **`run_mozzarellm` with `dry_run=True`** against the pinned mozzarellm: bundles built (UniProt accessions assigned), prompts written to `mozzarellm/run1/prompts_used/`, cost estimates returned, the context copy written into the run directory, and the combo-scoped `DAPI (marker of Nuclei)` wording confirmed present in the assembled system prompt. The `max_workers` fallback was exercised here — the pinned package does not take it and the call did not crash.
- **`annotate_cluster_anndata`** round-tripped through `write_h5ad`/`read_h5ad` on a synthetic clustering: obs columns and the nested `uns` mapping survive, and the visualizer's reader rebuilds the page's cluster dict and gene table from them.
- Both context JSONs validate against `mozzarellm`'s `ScreenContext` schema.
- `ruff format` / `ruff check` clean over the whole repo with the repo's own config.

Not run: the full `small_test_analysis` suite. Nothing this PR changes is on that DAG — the module is inert without the config block and the identical-DAG dry-run above is the evidence — and a real annotation run needs a provider API key.


mat10d:
Follow-ons, deliberately not in this PR:

**1. The analysis side is not wired yet.** The marimo notebook (`12_analyze.py`, brieflow-analysis #91) still calls `screen_context_from_screen(screen, config, LEIDEN_RESOLUTION)` and writes nothing to disk. To drive this module it needs to, per selected clustering: call `screen_context_for_combo(screen, config, channel_combo, cell_class, leiden_resolution)`, write the JSON under `config/mozzarellm_context/`, and emit the row into `config/mozzarellm_combo.tsv` alongside the `mozzarellm` config block. Until that lands the table and the context JSONs are written by hand.

**2. brieflow-auto interview rows.** Every operator-facing key in the new block (`run_name`, `screen_name`, `model`, `mode`, `mcp`, `include_features`, `include_strength`, `n_features`, `fdr_threshold`, `max_tokens`, `max_workers`) needs an interview row, a spec-CSV row and a `config_tiers.yaml` entry, plus the combo-table columns. That is a plugin release, so it is its own change.

**3. `mozzarellm.dry_run=True` still constructs a client**, and `create_client` validates the provider key before anything is assembled — so a dry run fails with `ANTHROPIC_API_KEY not found` on a machine without a key, even though it never calls the model. Worth moving the key check behind the dry-run branch in mozzarellm; harmless here because the pipeline path never dry-runs.

**4. The `mozzarellm:` config block now carries two unrelated things** — this module's parameters and the visualizer's default-view keys (`cell_class`, `channel_combo`, `leiden_resolution`, see `docs/source/3.running_modules.md`). That is why the module gates on `combo_fp` rather than on the block existing. If it gets confusing, the visualizer's three keys could move under `mozzarellm.default_view:` in a later change; doing it here would have broken every screen that already sets them.

**5. `run_mozzarellm(screen, config, …)` accepts `{}` for both** when `screen_context` and `organism_id` are supplied, which is how the rule calls it. It reads oddly. Reordering those to keyword-only with `None` defaults would be cleaner, but the notebook calls the function positionally, so it is worth doing together with item 1 rather than on its own.

mat10d:
Review of `b1bf1ae` (orchestrator pass). The module matches the repo's conventions — `map_outputs`/`outputs_to_targets`, the compartment-path helpers, thin scripts over `lib` — and the dry-run evidence for the identical-DAG case is the right gate. Two things worth changing before this runs on a real screen.

**1. One unrecoverable cluster fails the whole panel.** `annotate_clusters.py` raises when `result["errors"]` is non-empty, so with `retries: 2` a panel where 929 of 930 clusters succeed still writes no outputs after three attempts, and `annotate_cluster_anndata` never runs. Resume makes the retries cheap, which is right for transient failures, but a persistently unparseable response on a single cluster should not cost the panel. Suggest a tolerance: fail only when the failure count exceeds a small threshold (config key, default something like 1% or a handful), and otherwise write the outputs and report the failed cluster ids loudly. The traces stay on disk either way, so a later targeted rerun can fill them in.

**2. No `runtime` in `resources`.** Nothing else in the repo sets it, so this is consistent, but these jobs are far longer-lived than any other rule: a genome-wide panel is roughly 930 clusters, and even at `max_workers: 8` that is about three hours in one job, against minutes for most rules. Screens whose slurm profile carries a short default wall time will have panels killed mid-run. Resume means a killed job loses nothing but the in-flight clusters, so this is not data loss, just churn. Worth an explicit `runtime` so the requirement is stated rather than inherited.

Minor: `run_mozzarellm(..., {}, {}, ...)` passes empty screen and config dicts positionally now that the context comes from the row's JSON. Reads oddly at the call site; keyword defaults would say "not needed here" more clearly.

Issue #280, opened 2026-09-15
zarr3: zarr_max_levels and the storage_options compressor are both ignored, so OME-Zarr output is always single-level and near-uncompressed


petercla0119:
## Summary

On `zarr3` (`85aaf5b`), every OME-Zarr store the pipeline writes has one resolution level and zarr's fallback codec. Both knobs that look like they control this are disconnected:

- `all.zarr_max_levels` in `config_zarr.yml` is read by no code.
- `storage_options["compressor"]` on `write_image_omezarr` is documented but never forwarded.

Raising `max_levels` is not a safe fix by itself, because the pyramid path underneath it can downsample the channel axis.

Environment: `zarr 3.1.6`, `ome-zarr 0.13.0` (the `pyproject.toml` pins).

## zarr_max_levels is dead config

```
$ git grep -n zarr_max_levels
tests/small_test_analysis/config/config_zarr.yml:10:  zarr_max_levels: 1  # Number of pyramid resolution levels (1 = no downsampling)
```

One hit, the YAML line itself. `convert_image.py:34` never passes it:

```python
save_image(image_array, snakemake.output[0], channel_names=channel_names)
```

`save_image` (`image_io.py:96`) and `write_image_omezarr` (`:171`) both default to `max_levels=1`, so setting the config to 5 changes nothing.

## The compressor hook is a no-op

`storage_options` is declared at `image_io.py:174` and documented at `:191`. Those are its only two occurrences in the tree; it never reaches `write_image`. Passing a real codec through it produces byte-identical output:

```python
write_image_omezarr(img, out_a, axes="CYX", max_levels=3)
write_image_omezarr(img, out_b, axes="CYX", max_levels=3,
                    storage_options={"compressor": BloscCodec(cname="zstd", clevel=5,
                                                              shuffle=BloscShuffle.bitshuffle)})
```
```
out_a = 256964 bytes
out_b = 256964 bytes
```

What lands on disk is the default codec:

```
[{"name": "bytes", ...}, {"name": "zstd", "configuration": {"level": 0, "checksum": false}}]
```

zstd at level 0, which is the floor. Measured on the small-test data, single level so the pyramid does not confound the ratio, every roundtrip verified lossless:

| codec | phenotype tile (184 MB) | SBS tile (14.4 MB) |
|---|---|---|
| default (zstd level 0) | 0.668 | 0.738 |
| `blosc-zstd-noshuffle:5` | 0.667 | 0.736 |
[...]

PR #281, opened 2026-09-15
fix(io): make the OME-Zarr pyramid and compression actually reachable

petercla0119:
Closes #280.

Draft: opened for review of the approach, not yet proposed for merge.

## What is broken

`all.zarr_max_levels` is read by no code, and the `storage_options["compressor"]` hook on `write_image_omezarr` is documented but never forwarded. Every store the pipeline writes therefore has one resolution level and zarr's default `zstd level 0`, whatever the config says.

Raising `max_levels` alone is not a safe fix, because `ome_zarr.writer.write_image` on the pinned stack (`zarr 3.1.6`, `ome-zarr 0.13.0`) drops the compressor on its dask branch, and on its numpy branch scales the channel axis with Y/X (3 to 2 to 1) while smoothing level 0.

## Changes by file

**`workflow/lib/shared/image_io.py`** (+133 / -17)

- Adds `DEFAULT_MAX_LEVELS = 5` and `DEFAULT_ZARR_COMPRESSION = "blosc-zstd-bitshuffle"`. These are call-site defaults for the preprocess convert path, not writer defaults.
- Adds `_make_compressor()`, parsing `blosc-<cname>[-<shuffle>][:<clevel>]`. Any c-blosc codec is selectable from config without editing the module; `BloscCodec` validates `cname` itself. `"none"`/`None` keeps zarr's default.
- Adds a `compression` parameter to `save_image` and `write_image_omezarr`. Both keep `max_levels=1` / `compression=None`, so the other ~30 `save_image` callers are unchanged.
- Removes the `storage_options` parameter, which no caller ever passed.
- Replaces the `write_image` call with an explicit loop: `scaler.resize_image` per level, `da.to_zarr` with the codec attached, then `write_multiscales_metadata`. Level 0 is written verbatim and the channel axis is never scaled.
- Caps the c-blosc thread pool via `BRIEFLOW_BLOSC_NTHREADS` (default 4) so tile conversion does not spawn one thread per core on top of Snakemake's parallelism.

**`workflow/scripts/preprocess/convert_image.py`** (+16 / -2)

Reads `all.zarr_max_levels` and `all.zarr_compression` and passes them to `save_image`, falling back to the two constants. This is the only consumer of them.

**`tests/small_test_analysis/config/config.yml`** (+4) and **`config_zarr.yml`** (+3 / -1)

Document both keys. `zarr_max_levels` goes 1 to 5 in the zarr config; `zarr_compression` is new.

**`tests/test_omezarr.py`** (+240 / -100)

- `TestZarrStructural` resolved stores by rebuilding `preprocess/images/sbs/P-..__image.zarr`, a TIFF-style filename with a `.zarr` suffix that exists in neither output mode, so all five of its tests skipped. They now locate image groups by searching for `ome.multiscales` metadata, which follows the HCS plate layout zarr mode actually writes.
- `test_zarr_compression_applied` asserted `ratio < 5.0`, which uncompressed 16-bit data passes at roughly 1.0. Replaced by `test_zarr_compression_codec_is_configured`, asserting the codec by name on every level.
- Adds `test_level0_full_resolution_and_channels_preserved`, which fails if C is scaled across levels or level 0 is not full resolution.
- Adds `TestCompressionAndPyramid`: pyramid shape, level-0 bit-identity, label nearest-neighbour, codec reaching every level, five parametrized codec specs each verified lossless, malformed specs raising, and conservative writer defaults.

No rules, targets, or other modules are touched.

## Validation

Ran the real pipeline on this branch, small-test data, zarr mode:

```
snakemake --cores 8 --configfile config/config.yml config/config_zarr.yml --until all_preprocess
166 of 166 steps (100%) done
```

The converted store now carries what the config asks for:

```
levels declared: ['0', '1', '2', '3', '4']
  level 0: shape=(1, 5, 1, 1200, 1200) codecs=['bytes', 'blosc']
  level 1: shape=(1, 5, 1,  600,  600) codecs=['bytes', 'blosc']
  level 2: shape=(1, 5, 1,  300,  300) codecs=['bytes', 'blosc']
  level 3: shape=(1, 5, 1,  150,  150) codecs=['bytes', 'blosc']
  level 4: shape=(1, 5, 1,   75,   75) codecs=['bytes', 'blosc']
level0 blosc: {'cname': 'zstd', 'clevel': 5, 'shuffle': 'bitshuffle', 'typesize': 2}
```

C stays at 5 across every level and Y/X halve, which is the behaviour the numpy branch of `write_image` breaks.

Test suite against that run:

```
37 passed, 1 skipped
```

The skip is `test_zarr_tiff_equivalence_sbs`, which needs a TIFF run to compare against. Before this PR all five `TestZarrStructural` tests skipped; four now execute.

To confirm the tests can fail, I built a store with the exact defect from #280 (single level, writer defaults, no codec) and pointed them at it:

```
FAILED test_zarr_compression_codec_is_configured
FAILED test_omezarr_multiscale_structure
2 failed, 2 passed, 1 skipped
```

The previous versions skipped silently on that same input.

One thing the end-to-end run caught that unit tests did not: my first version of the discovery helper searched all of `preprocess/` and sorted `ic_fields/` first, so it asserted a codec against illumination-correction stores. Those go through `calculate_ic_*` with the conservative writer defaults and are correctly single-level and uncompressed. The helper is now scoped to `preprocess/{sbs,phenotype}`, the convert output that actually threads the config (774703e).

### Measured compression

Real small-test tiles, single level so the pyramid does not confound the ratio, every roundtrip verified lossless:

| codec | phenotype tile (184 MB) | SBS tile (14.4 MB) |
|---|---|---|
| default (zstd level 0) | 0.668 | 0.738 |
| `blosc-zstd-noshuffle:5` | 0.667 | 0.736 |
| `blosc-zstd-bitshuffle:5` | 0.570 | 0.651 |
| `blosc-zstd-shuffle:5` | 0.554 | 0.629 |

Roughly 15% smaller than current output. Most of the gain is the shuffle filter rather than the zstd level, since `noshuffle` lands on the default. `clevel 9` buys another 2 to 3% for about 9x the write time, so it is not a sensible default.

Pyramid cost on the 184 MB tile: 33.3% storage and about 11s per tile over single level. Levels 4 and 5 are the poor end, 2.1% storage for 6.5s.

### Environment note

Two `TestND2Metadata` tests fail without `imaris-ims-file-reader` installed. That reproduces identically on pristine `zarr3`; the package is declared in `pyproject.toml` but was missing from my env. Unrelated to this change, and they pass once it is installed.

## Open questions for maintainers

1. **Default codec.** I kept `blosc-zstd-bitshuffle` because it is the right choice for 12-bit-in-16 sensor data, which is common. Byte shuffle measured better on both of my tiles because that camera fills 15 of 16 bits, leaving no dead bit planes for bitshuffle to exploit. Happy to switch the shipped default if you would rather optimize for the small-test data.
2. **Pyramid depth.** 5 is what the config comment implied was intended. 3 captures nearly the same storage profile for half the write time.
3. **Scope.** This threads config into the preprocess convert path only. The SBS/phenotype tile writers and `hcs.py` finalize still use writer defaults, so they remain uncompressed and single-level. Happy to extend here or leave it to a follow-up.


petercla0119:
Superseded by #288 — same head commit, no content change, still a draft. Head branch renamed to `up/fix-omezarr-pyramid-compression` so upstream-bound branches are distinguishable from this fork's local development branches. Closing here; follow #288.

